PR #12, opened 2017-05-22
observe field attributes in group members

ralreiroe:


lutnos:
case class MES isn't used?

PR #16, opened 2017-06-05
Eeitt 995 grouphoriz

ralreiroe:


app/uk/gov/hmrc/gform/models/FieldValueRaw.scala
```diff
@@ -136,11 +136,19 @@ case class FieldValueRaw(
...
   } yield Date(f, o, v)
 
-  private lazy val groupOpt: Opt[Group] = fields.fold(groupOptErr)(optGroup)
-  private lazy val groupOptErr: Opt[Group] = InvalidState(s"""Require 'fields' element in Group""").asLeft
-  private def optGroup(rawFields: List[FieldValueRaw]): Opt[Group] = rawFields.map(_.getFieldValue()).partition(_.isRight) match {
-    case (ueorfvs, Nil) => Group(ueorfvs.map(_.right.get)).asRight
-    case (_, ueorfvs) => ueorfvs.map(_.left.get).head.asLeft
+  private lazy val groupOpt: Either[UnexpectedState, Group] = fields.fold(noRawFields)(createGroupFromRawFieldsOrReturnUnexpectedState)
+  private lazy val noRawFields: Either[UnexpectedState, Group] = InvalidState(s"""Require 'fields' element in Group""").asLeft
+  private def createGroupFromRawFieldsOrReturnUnexpectedState(rawFields: List[FieldValueRaw]): Either[UnexpectedState, Group] = {
```
pawelpanasewicz: The previous names and return types were neat. I don't really understand if we really need to put extra long names and types here?

Let's consider this name for example:
`private def createGroupFromRawFieldsOrReturnUnexpectedState(rawFields: List[FieldValueRaw]): Either[UnexpectedState, Group]`

`create` - we define a function here so it returns/creates something. This word is unnecessary.
`Group` - the function returns a `Group` thus it's obvious that the something we create here is a group thus this word is unnecessary.
`fromRawFields` - the same here: function takes list of raw fields (we can easy spot it reading the type signature)
`OrReturnUnexpectedState` - and the same here, function returns `Opt[Something]` - which means that it returns something or in case of failure `UnexpectedState`. 
In my opinion above words are obvious and don't require extra verbosity. 
Could we please not use such naming?

lutnos: Yes, I'm inclined to agree with Pawel here, shorter names are better

app/uk/gov/hmrc/gform/models/FieldValueRaw.scala
```diff
@@ -65,7 +65,7 @@ case class FieldValueRaw(
 
   def toFieldValue = Reads[FieldValue] { _ => getFieldValue fold (us => JsError(us.toString), fv => JsSuccess(fv)) }
 
-  private def getFieldValue(): Opt[FieldValue] = optMES.flatMap(mes => optComponentType.map(ct => mkFieldValue(mes, ct)))
+  private def getFieldValue(): Either[UnexpectedState, FieldValue] = optMES.flatMap(mes => fromRawToFinishedComponentType.map(ct => mkFieldValue(mes, ct)))
```
lutnos: @ralreiroe What is the reason for changing from Opts to Eithers?

ralreiroe: Yes, Opt is shorter but it saves me looking up what it is...

app/uk/gov/hmrc/gform/models/FieldValueRaw.scala
ralreiroe: I guess the two sides of the discussion is:
(a) fields.fold(groupOptErr)(optGroup) - requires you to look at the types and the signatures elsewhere to understand what's going on.
(b) fields.fold(noRawFields)(createGroupFromRawFieldsOrReturnUnexpectedState) - requires less of that. Although using the shorter fields.fold(noRawFields)(groupFromRawFieldsOrUnexpectedState) would achieve similar.

Plus, one can observe that once one is very familiar with the types and signatures and the code base in general, the requirement goes away.

app/uk/gov/hmrc/gform/models/FieldValueRaw.scala
pawelpanasewicz: I know that it's pain for the first time when you see it but once you remember it it saves much more time. Especially that it's widely used in this project :)

PR #18, opened 2017-06-05
Pawel/common spec

pawelpanasewicz:
+118 −157


pawelpanasewicz:
+118 −157

test/uk/gov/hmrc/gform/services/FormServiceSpec.scala
```diff
@@ -16,19 +16,16 @@
...
 import play.api.libs.json.{ JsObject, Json }
-import uk.gov.hmrc.gform.{ FindOneCheck, TypeclassFixtures }
 import uk.gov.hmrc.gform.exceptions.InvalidState
 import uk.gov.hmrc.gform.models._
 import uk.gov.hmrc.gform.typeclasses.{ FindOne, Insert, Update }
+import uk.gov.hmrc.gform.{ FindOneCheck, Spec, TypeclassFixtures }
 
 import scala.concurrent.ExecutionContext.Implicits.global
 
-class FormServiceSpec extends FlatSpec with Matchers with TypeclassFixtures with ScalaFutures with EitherValues with MockFactory {
```
ralreiroe: Sometimes in the past I have found it useful to know which particular traits were mixed in for the test at hand.

lutnos: Yes, me too.  But I can see the advantage of the brevity & if occassionaly necessary, when investigating something, it is usually little trouble to revert temporarily to mining in whatever is actually needed.

PR #28, opened 2017-06-19
Eeitt 1078 - Added a few fixes related to address

tenochdlc:


pawelpanasewicz:
Looks good. We must only follow the same steps for other components to be more consistent.

PR #47, opened 2017-07-06
Implemented delete template, will ask whether it requires unit tests

HarrisonBaxter:


app/uk/gov/hmrc/gform/controllers/FormTemplates.scala
```diff
@@ -103,7 +103,15 @@ class FormTemplates()(
       }
   }
 
-  def delete(formTypeId: FormTypeId, version: Version) = Action.async { implicit request =>
-    Future.successful(NotImplemented)
+  def delete(formTypeId: FormTypeId, version: String) = Action.async { implicit request =>
```
pawelpanasewicz: @ralreiroe Do we need to delete by version and formTypeId? This would delete all forms for all users for given version and type?
How about deleting particular form designated by only formId?

HarrisonBaxter: The forms are stored in a separate collection, deleting the template means you can't access the form but the data is not deleted from the forms collection

pawelpanasewicz: Oh, I didn't see the "Template" word :P All good now

PR #70, opened 2017-07-21
adding in logging to show full form to better debug issues with save4…

DanielConnelly:
…Later


lutnos:
I suggested this logging to Daniel.

When we save to Mongo, it is possible to later query to find out what is present and what is in it.  Because save4later encrypts the data, it can be an obstacle.  A good solution would be to have a way to disable encryption locally, but for that we would need a change to the save4later library.

Logging would enable us to search the log file for what has been written to save4later.

DanielConnelly:
Is this Ok to merge?

pawelpanasewicz:
Ohh, I see now. Let's have them for now. Later we can provide switch in save4later which additionally will save the same records in mongo. Switch will be enabled only in local. This will allow us to inspect what has been saved.

PR #73, opened 2017-07-24
Fix for Json Validation bug in repeatsMax field

tenochdlc:


app/uk/gov/hmrc/gform/models/FieldValueMaker.scala
```diff
@@ -60,7 +60,7 @@ class FieldValueMaker(json: JsValue) {
   lazy val infoText: Option[String] = (json \ "infoText").asOpt[String]
   lazy val infoType: Option[String] = (json \ "infoType").asOpt[String]
   lazy val shortName: Option[String] = (json \ "shortName").asOpt[String]
-  lazy val repeatsMax: Option[Int] = (json \ "repeatsMax").asOpt[Int]
+  lazy val repeatsMax: JsResult[Option[Int]] = (json \ "repeatsMax").validateOpt[Int]
```
pawelpanasewicz: We should convert x to Either at this stage or else we wouldn't be able to rely on for comprehension in other places when computing dependent values.
More over I think we should convert all other fields from `Option` to `Opt[T]` as well and keep staing consistent in way we code.
^ @ralreiroe

PR #74, opened 2017-07-25
min/max/presentationHint

ralreiroe:


app/uk/gov/hmrc/gform/models/FieldValueMaker.scala
```diff
@@ -44,6 +44,13 @@ class FieldValueMaker(json: JsValue) {
     format.map(FormatParser.validate).sequence
   }
 
+  lazy val presentationHint: Option[String] = (json \ "presentationHint").asOpt[String]
```
pawelpanasewicz: For such json `{"presentationHint" : 123}` presentation hint will become `None`. What we need is to have`Left("expected String but got int")`.

app/uk/gov/hmrc/gform/models/FieldValueMaker.scala
```diff
@@ -60,22 +67,28 @@ class FieldValueMaker(json: JsValue) {
   lazy val infoText: Option[String] = (json \ "infoText").asOpt[String]
   lazy val infoType: Option[String] = (json \ "infoType").asOpt[String]
   lazy val shortName: Option[String] = (json \ "shortName").asOpt[String]
-  lazy val repeatsMax: JsResult[Option[Int]] = (json \ "repeatsMax").validateOpt[Int]
-  lazy val repeatsMin: Option[Int] = (json \ "repeatsMin").asOpt[Int]
+  lazy val jsResultMaybeRepeatsMax: JsResult[Option[Int]] = (json \ "repeatsMax").validateOpt[Int]
+  lazy val jsResultMaybeRepeatsMin: JsResult[Option[Int]] = (json \ "repeatsMin").validateOpt[Int]
```
pawelpanasewicz: Here we could return `Opt[...]` instead of `JsResult` because later is not operable in for comprehension toghether with `Either`. Other solution is to convert all other fields to be `JsResult` rather than `Either`.

app/uk/gov/hmrc/gform/models/FieldValueMaker.scala
```diff
@@ -60,22 +67,28 @@ class FieldValueMaker(json: JsValue) {
   lazy val infoText: Option[String] = (json \ "infoText").asOpt[String]
   lazy val infoType: Option[String] = (json \ "infoType").asOpt[String]
   lazy val shortName: Option[String] = (json \ "shortName").asOpt[String]
-  lazy val repeatsMax: JsResult[Option[Int]] = (json \ "repeatsMax").validateOpt[Int]
-  lazy val repeatsMin: Option[Int] = (json \ "repeatsMin").asOpt[Int]
```
tenochdlc: There are many ways to skin a cat :-), I think this solution and mine work fine. I don't see the need to have the Opt at this level, but if kittens could die if we don't have Opt at this level, here's the other part of my idea:

lazy val repeatsMax: Opt[Option[Int]] = handleJsonValidationResult( (json \ "repeatsMax").validateOpt[Int] )
lazy val repeatsMin: Opt[Option[Int]] = handleJsonValidationResult( (json \ "repeatsMin").validateOpt[Int] )

app/uk/gov/hmrc/gform/models/FieldValueMaker.scala
```diff
@@ -202,29 +215,30 @@ class FieldValueMaker(json: JsValue) {
     for {
       fieldValues <- fieldValuesOpt.right
       format <- optMaybeFormatExpr.right
-      group <- validateAndBuildGroupField(fieldValues, orientation(format))
+      repMax <- jsResultToOpt(jsResultMaybeRepeatsMax)
```
pawelpanasewicz: see above comment about converting about `Either` vs `JsResult`

app/uk/gov/hmrc/gform/models/FieldValueMaker.scala
pawelpanasewicz: > I don't see the need to have the Opt at this level

@tenochdlc How about the need to have consistent code?

ralreiroe:
see new commits...

ralreiroe:
merging after final changes and review with Pawel

PR #117, opened 2017-08-22
utility for uploading templates

pawelpanasewicz:


it/uk/gov/hmrc/gform/TemplatesInSampleDataSpec.scala
```diff
@@ -4,23 +4,45 @@ import java.nio.file.{Files, Paths}
...
     }
+}
 
+object TemplatesInSampleDataSpec {
   lazy val templates = Files
     .newDirectoryStream(Paths.get("sample-data"))
     .asScala
     .filter(Files.isRegularFile(_))
-    .filter(_.getFileName.toString.matches("template-add(.*)json"))
+    .filter(_.getFileName.toString.matches("template-aaa(.*)json"))
```
pawelpanasewicz: after changing filter `template-aaa` to `template` we could test all templates

PR #130, opened 2017-08-29
Daniel/validators work

DanielConnelly:


app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import com.typesafe.config.Config
+import play.api.libs.json.{ JsValue, Json, OFormat }
+import uk.gov.hmrc.gform.wshttp.WSHttp
+import uk.gov.hmrc.play.http.logging.Authorization
+import uk.gov.hmrc.play.http.{ HeaderCarrier, NotFoundException }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class DesConnector(wSHttp: WSHttp, baseUrl: String, desConfig: Config) { //TODO make the connector easy to move from service to service
```
pawelpanasewicz: 1. How about passing instance of DesConnectorConfig here instead of typesafe config? Typesafe config might or might not have all required fields.

```
case class DesConnectorConfig(baseUrl: String, basePath: String, token: String, environment: String) 
```

2. Since it's in proper package is easily removable, so we don't need this comment do we? `//TODO make the connector easy to move from service to service`

app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    wSHttp.POST[JsValue, AddressDes](s"$baseUrl${desConfig.getString("base-path")}/registration/organisation/utr/$utr", lookupJson)
+      .map(compare)
+      .recover {
+        case _: NotFoundException => false
+      }
+  }
+
+  def createHC =
```
pawelpanasewicz: Looks like the `implicit` keyword is missing or we don't use this method.

app/uk/gov/hmrc/gform/fileupload/FileUploadModule.scala
```diff
@@ -20,7 +20,7 @@ import uk.gov.hmrc.gform.config.ConfigModule
 import uk.gov.hmrc.gform.time.{ TimeModule, TimeProvider }
 import uk.gov.hmrc.gform.wshttp.WSHttpModule
 
-class FileUploadModule(configModule: ConfigModule, wSHttpModule: WSHttpModule, timeModule: TimeModule) {
+class FileUploadModule(configModule: ConfigModule, val wSHttpModule: WSHttpModule, timeModule: TimeModule) {
```
pawelpanasewicz: Let's remove this `val` keyword. We don't want to expose sub-modules from module. Module suppose to be responsible only for creating instances.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
```diff
@@ -0,0 +1,114 @@
...
+import uk.gov.hmrc.gform.core.parsers.ExprParsers
+import uk.gov.hmrc.play.http.HeaderCarrier
+import play.api.libs.functional.syntax._
+import play.api.libs.functional.syntax._
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+sealed trait Validator {
+  def errorMessage: String
+  def validate(data: Map[FieldId, Seq[String]])(f: ((String, String)) => Future[Boolean])(implicit ec: ExecutionContext, hc: HeaderCarrier): Future[Validated[Map[FieldId, Set[String]], Unit]]
```
pawelpanasewicz: Here we're breaking _Single Responsibility Principle_. `Validator` suppose to be a part of a template, just plain data structure, kind of instruction for later interpretation. Interpretation and therefore validating  should happen somwhere else. 
How about moving this method to somewhere else (maybe to `ValidationService`)?

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
```diff
@@ -0,0 +1,114 @@
...
+
+}
+
+case class HMRCUTRPostcodeCheckValidator(errorMessage: String, utr: FormCtx, postcode: FormCtx) extends Validator {
+
+  val utrFieldId = FieldId(utr.value)
+  val postcodeFieldId = FieldId(postcode.value)
+
+  def validate(data: Map[FieldId, Seq[String]])(f: ((String, String)) => Future[Boolean])(implicit ec: ExecutionContext, hc: HeaderCarrier): Future[Validated[Map[FieldId, Set[String]], Unit]] = {
+    val dataGetter: FieldId => String = id => data.get(id).toList.flatten.headOption.getOrElse("")
```
pawelpanasewicz: hmm, if there is no data for given id maybe we should immediately return invalid?

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
```diff
@@ -0,0 +1,114 @@
...
+  val writes: OWrites[HMRCUTRPostcodeCheckValidator] = writesCustom
+  val readCustom: Reads[HMRCUTRPostcodeCheckValidator] = ((JsPath \ "errorMessage").read[String] and
+    (JsPath \ "parameters" \\ "utr").read[FormCtx] and
+    (JsPath \ "parameters" \\ "postcode").read[FormCtx])(HMRCUTRPostcodeCheckValidator.apply _)
+
+  val reads = readCustom | (basic: Reads[HMRCUTRPostcodeCheckValidator])
+  implicit val format: OFormat[HMRCUTRPostcodeCheckValidator] = OFormat(reads, writesCustom)
+}
+
+//case class Validators(validatorName: String, errorMessage: String, parameters: Map[String, FormCtx]) {
```
pawelpanasewicz: Let's get rid of these comments

app/uk/gov/hmrc/gform/validation/ValidationController.scala
```diff
@@ -0,0 +1,42 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class ValidationController(desConnector: DesConnector) extends BaseController {
+
+  def validateAddressAtDes(utr: String, postcode: String) = Action.async { implicit request =>
+    desConnector.lookup(utr, postcode).map(if (_) NoContent else NotFound)
+  }
+
+  def testValidatorStub(utr: String) = Action.async { implicit request =>
```
pawelpanasewicz: I'm not sure if this test logic should land in `'TestOnlyController`. To discuss.

app/uk/gov/hmrc/gform/validation/ValidationController.scala
```diff
@@ -0,0 +1,42 @@
...
+    Logger.debug(s"this it the UTR that is passed to the function ${utr}")
+    if (utr.startsWith("1")) {
+      Logger.debug("it hit here")
```
pawelpanasewicz: Let's get rid of such comments :)

tst123.tmp.json
```diff
@@ -0,0 +1,272 @@
+{
```
pawelpanasewicz: Opps, we forgot to remove this file

tst123raw.tmp.json
```diff
@@ -0,0 +1,114 @@
+{
```
pawelpanasewicz: this as well should be removed

pawelpanasewicz:
Well done, however still few minor comments.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
DanielConnelly: thats a confusing one, as the postcode can be blank ... but cannot be optional I was speaking to Nathan about this and its confusing.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
DanielConnelly: I thought about that. however it is called inside the frontend validation service and the desired function is passed in. its cause each validator may have different implementation depending on what needs to be done on it. and it makes more sense to me to add it to this object.

app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -0,0 +1,58 @@
...
+       "isAnAgent": false
+      }""") //TODO add in actual regime we are looking for
+
+  def lookup(utr: String, postCode: String)(implicit hc: HeaderCarrier, ex: ExecutionContext): Future[Boolean] = {
+
+    def compare(address: AddressDes) = {
+      address.postalCode.replace(" ", "").equalsIgnoreCase(postCode.replace(" ", "")) || address.postalCode == "Valid"
+    }
+
+    implicit val hc = HeaderCarrier(extraHeaders = Seq("Environment" -> desConfig.environment), authorization = Some(Authorization(desConfig.authorizationToken)))
```
pawelpanasewicz: todo: we need merge this `hc` with HeaderCarrier made from request.

PR #139, opened 2017-09-04
More than 2 values supported [EEITT-1261]

DanielConnelly:


test/uk/gov/hmrc/gform/wshttp/StubbedWSHttp.scala
```diff
@@ -40,11 +40,12 @@ class StubbedWSHttp(response: HttpResponse) extends WSHttp {
 //If you want to use WSHttp outside play app you must provide your WSClient. Otherwise it blows up.
 //See https://github.com/hmrc/http-verbs/issues/60
 //Don't use it on production ('ws.close()' logic is missing)
+/*
```
pawelpanasewicz: This should be fixed so after rebase you would not need to comment this code

PR #167, opened 2017-09-22
making correct json obj

DanielConnelly:


app/uk/gov/hmrc/gform/bank_account_reputation/BankAccountReputationConnector.scala
```diff
@@ -43,7 +43,7 @@ object Account {
   val basic: OFormat[Account] = Json.format[Account]
 
   val writes: OWrites[Account] = OWrites[Account] { o =>
-    Json.obj("account" -> s"""{sortCode: ${o.sortCode}, accountNumber: ${o.accountNumber}}""")
+    Json.obj("account" -> Json.obj("sortCode" ->  s"${o.sortCode}", "accountNumber" ->  s"${o.accountNumber}"))
```
lutnos: It would be simpler to put s"${o.sortCode}" as o.sortCode  & likewise accountNumber

PR #220, opened 2018-01-25
[CC][GFC-99]Added in developmentPhase to json struture

cavster:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormTemplate.scala
```diff
@@ -23,6 +23,7 @@ case class FormTemplate(
   _id: FormTemplateId,
   formName: String,
   description: String,
+  developmentPhase: Option[String] = None,
```
lutnos: I don't think this should be an option of string.

This allows the user to enter just anything for developmentPhase, anything can get displayed on the browser and the View starts matching particular values, e.g. "live".

developmentPhase ought to be an enumeration implemented with a sealed trait, in a similar way to FormCategory .

That way only acceptable value will be parsed when the template is uploaded.

It would also be nice to encapsulate the implementation of default valueuk.gov.hmrc.gform.summary.SummaryRenderingService#renderSummary is one way, but there may be a better one

PR #264, opened 2018-08-22
GFC-557 Check status from pdf generator & wait for response from email service

lutnos:


app/uk/gov/hmrc/gform/pdfgenerator/PdfGeneratorConnector.scala
```diff
@@ -30,8 +30,16 @@ class PdfGeneratorConnector(servicesConfig: ServicesConfig, wSHttp: WSHttp) {
   def generatePDF(payload: Map[String, Seq[String]], headers: Seq[(String, String)])(
     implicit hc: HeaderCarrier): Future[Array[Byte]] = {
     Logger.info(s"generate pdf, ${loggingHelpers.cleanHeaderCarrierHeader(hc)}")
-    wSHttp.buildRequest(s"$baseURL/pdf-generator-service/generate").withHeaders(headers: _*).post(payload).map {
-      _.bodyAsBytes.toArray
+    val url = s"$baseURL/pdf-generator-service/generate"
+    wSHttp.buildRequest(url).withHeaders(headers: _*).post(payload).flatMap { response =>
+      {
+        val status = response.status
+        if (status <= 200 && status < 300) {
```
josef-vlach: Is this check correct? Don't we want to use something like [is2xx](https://github.com/hmrc/http-core/blob/e7dc6fc31454f9fd9ceca18dacef5828fa5a6ceb/src/main/scala/uk/gov/hmrc/http/HttpErrorFunctions.scala#L46) from http-core ?

PR #269, opened 2018-10-05
Dave/gfc 614/use draft retrieval method

lutnos:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DraftRetrievalMethod.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    override def reads(json: JsValue): JsResult[DraftRetrievalMethod] =
+      json match {
+        case JsString("onePerUser") =>
+          JsSuccess(OnePerUser)
+        case JsString("formAccessCodeForAgents") =>
+          JsSuccess(FormAccessCodeForAgents)
+        case JsString(err) =>
+          JsError(s"only two valid draft retrieval methods: onePerUser, or formAccessCodeForAgents; $err is not valid")
+        case _ => JsError("Failure")
```
josef-vlach: Better description what went wrong would be useful. Any indication about DraftRetrievalMethod being wrong will be time-saver.

lutnos: Thanks, yes, is this now better?

josef-vlach: I'm concerned actually about this line:

```
case _ => JsError("Failure")
```

PR #274, opened 2018-10-17
Enables parsing of empty strings

WitchellJack:


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -82,6 +78,8 @@ object ValueParser {
     | "auth" ~ "." ~ authInfo ^^ { (loc, _, _, authInfo) =>
       AuthCtx(authInfo)
     }
+    | quotedConstant
```
lutnos: Oops, this is the other one!

lutnos: My mistake!

PR #291, opened 2019-01-14
David/698/send email parameters 2

Davidrymer97:


test/uk/gov/hmrc/gform/sharedmodel/ExampleData.scala
```diff
@@ -298,6 +298,8 @@ trait ExampleFormTemplate {
 
   def emailTemplateId = "test-email-template-id"
 
+  def emailParameters = List(EmailParameter("fullName", "john smith"), EmailParameter("email", "test@test.com"))
```
lutnos: This suggests that the email parameter values are literal strings, but actually they need to be references to form field values, see https://confluence.tools.tax.service.gov.uk/display/DF/05.+Gforms+Specification

app/uk/gov/hmrc/gform/submission/SubmissionService.scala
```diff
@@ -228,4 +230,13 @@ object SubmissionServiceHelper {
     sectionsToSubmit.traverse(toSectionFormField)
   }
 
+  def getEmailParameterValues(formTemplate: FormTemplate, form: Form): Map[String, String] =
+    formTemplate.emailParameters
+      .flatMap(
+        parameter =>
+          form.formData.fields
+            .find(field => field.id.value == parameter.emailTemplateVariable)
```
lutnos: This looks up field values using the emailTemplateVariable, but that read ${xxx}, but you need to search for xxx not ${xxx}, see https://confluence.tools.tax.service.gov.uk/display/DF/05.+Gforms+Specification

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/EmailParameter.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.gform.sharedmodel.formtemplate
+
+import play.api.libs.json.{ Json, OFormat }
+case class EmailParameter(emailTemplateVariable: String, value: String)
```
lutnos: The variables in the email parameters are of the form ${xxx}, you need to check that is true and strip out the ${} .

It would also be a good idea to check when the template is parsed that the field xxx actually exists.  Otherwise no problem will be found until run time & even then only in QA or Production where email sending works

josef-vlach: 👍 for checking that fields actually exists, but I'm not sure we want simple strip out `${}`. Shouldn't we treat `value` as full blown expression (`Expr`)?

PR #294, opened 2019-01-17
David/698/send email parameters

Davidrymer97:


test/uk/gov/hmrc/gform/core/parsers/ValueParserSpec.scala
```diff
@@ -23,6 +23,8 @@ import uk.gov.hmrc.gform.formtemplate._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.destinations.Destinations
 
+import scala.collection.immutable.List
```
lutnos: scala.collection.immutable.List is included in scala.Predef which is available in every scala source file by default, so there is no need to explicitly import it

PR #295, opened 2019-01-17
Daryl/653/value rounding

DarylMConway:


DarylMConway:
found further complications with the given branch as javascript has different rounding modes as well as rounding errors compared to the scala version. closing pull request as a result

PR #297, opened 2019-01-18
David/574/implement date formats using first day and last day

Davidrymer97:


app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
```diff
@@ -57,11 +57,23 @@ object BasicParsers {
       fn(month, day)
     })
 
+  def firstOrLastDay[A](string: String, fn: (String, String) => A): Parser[A] =
+    yearMonth ~ string ^^ { (loc, year, month, _) =>
+      fn(year.toString, month.toString)
+    } | "YYYY" ~ delimiter ~ "MM" ~ delimiter ~ string ^^ { (loc, _, _, _, _, _) =>
```
josef-vlach: I do not know why this PR is first one to deal with validation contained in note 

> Note: "YYYY", "MM" and "DD" match any year, month or date; alternatively any of these can be replaced by an actual year, month or date in order to provide a mask e.g. YYYY-04-DD for any day in April.

but since it tries to do that we need to make sure we implement it properly and across all existing cases ie. `YYYY-MM-DD | next-MM-DD | previous-MM-DD | YYYY-MM-firstDay | YYYY-MM-lastDay`

lutnos:
@PasGat is still working on this

PR #301, opened 2019-01-22
GFC-736: added in rounding mode ADT for input and altered text compon…

DarylMConway:
…ent to have the correct number of inputs

### DIFF
--- a/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
+++ b/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
@@ -26,11 +26,7 @@ import scala.collection.immutable._
 
 sealed trait ComponentType
 
-case class Text(
-  constraint: TextConstraint,
-  value: Expr,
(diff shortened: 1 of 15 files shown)


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
```diff
@@ -26,7 +26,11 @@ import scala.collection.immutable._
 
 sealed trait ComponentType
 
-case class Text(constraint: TextConstraint, value: Expr, displayWidth: DisplayWidth = DisplayWidth.DEFAULT)
+case class Text(
+  constraint: TextConstraint,
+  value: Expr,
+  displayWidth: DisplayWidth = DisplayWidth.DEFAULT,
+  roundingMode: RoundingMode = RoundingMode.defaultRoundingMode)
```
josef-vlach: Why is `roundingMode ` here? 

Text should have no knowledge of rounding. That knowledge should be responsibility of constraint, namely `Number` etc. And as I can see you've also added it there.

PR #308, opened 2019-01-28
Change emailParameters to Optional in form template

Davidrymer97:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormTemplate.scala
```diff
@@ -39,7 +39,7 @@ case class FormTemplate(
   destinations: Destinations,
   authConfig: formtemplate.AuthConfig,
   emailTemplateId: String,
-  emailParameters: List[EmailParameter],
+  emailParameters: Option[List[EmailParameter]],
```
VlachJosef: Having `Option[List[EmailParameter]` seems ambiguous. What is difference between `None` value and `Some(List.empty)` value?

Much better representation would be `Option[NonEmptyList[EmailParameter]]`.

josef-vlach: Having `Option[List[EmailParameter]` seems ambiguous. What is difference between `None` value and `Some(List.empty)` value?

Much better representation would be `Option[NonEmptyList[EmailParameter]]`.

PR #310, opened 2019-01-29
Added submissionReference to parser

kylemchownhmrc:


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -72,6 +72,9 @@ object ValueParser {
     | "user" ~ "." ~ userField ^^ { (loc, _, _, userField) =>
       UserCtx(userField)
     }
+    | "form" ~ "." ~ "submissionReference" ^^ { (loc, _, _, fieldName) =>
```
lancewalton: We need to have unit tests to make sure this bit of parsing does the right thing

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
```diff
@@ -45,6 +45,7 @@ final case class EeittCtx(value: Eeitt) extends Expr
 final case class UserCtx(value: UserField) extends Expr
 final case class Constant(value: String) extends Expr
 final case object Value extends Expr
+final case object SubmissionReference extends Expr
```
lancewalton: Extend ExprGen to include this case so that it's included in the JSON round tripping tests.

PR #318, opened 2019-02-04
GFC-756: PDF Submission to FileUpload does not handle failure

lancewalton:
It does now.

### DIFF
--- a/app/uk/gov/hmrc/gform/dms/DmsSubmissionController.scala
+++ b/app/uk/gov/hmrc/gform/dms/DmsSubmissionController.scala
@@ -19,10 +19,15 @@ package uk.gov.hmrc.gform.dms
 import java.time.{ Clock, LocalDateTime }
 import java.util.Base64
 
+import cats.data.EitherT
+import cats.instances.future._
 import org.apache.pdfbox.pdmodel.PDDocument
 import play.api.libs.json.JsValue
-import play.api.mvc.Action
+import play.api.Logger
+import play.api.mvc.{ Action, Result }
 import uk.gov.hmrc.gform.controllers.BaseController
+import uk.gov.hmrc.gform.core.fromFutureA
+import uk.gov.hmrc.gform.exceptions.UnexpectedState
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/gform/dms/DmsSubmissionController.scala
```diff
@@ -67,10 +73,15 @@ class DmsSubmissionController(
...
           metadata.businessArea)
-        _ <- fileUpload.submitEnvelope(submission, summaries, dmsSubmission, 0)
+        _ <- fromFutureA(fileUpload.submitEnvelope(submission, summaries, dmsSubmission, 0))
       } yield {
         NoContent
       }
+
+      result.leftMap { l =>
+        Logger.info(s"Problem submitting to DMS: ${l.error}")
+        InternalServerError
```
lutnos: Hmm, I am a bit worried by this.

Normally such errors are handled in ErrorHandler, according to the upstream error.

I would expect errors from the file upload to be handled in the same way, albeit that you might want to additionally log here.

app/uk/gov/hmrc/gform/pdfgenerator/PdfGeneratorConnector.scala
```diff
@@ -18,26 +18,27 @@ package uk.gov.hmrc.gform.connectors
...
-      {
+    fromFutureOptA {
+      wSHttp.buildRequest(url).withHeaders(headers: _*).post(payload).map { response =>
         val status = response.status
         if (status >= 200 && status < 300) {
-          Future.successful(response.bodyAsBytes.toArray)
+          Right(response.bodyAsBytes.toArray)
         } else {
-          Future.failed(new Exception(s"POST to $url failed with status $status. Response body: '${response.body}'"))
+          Left(UnexpectedState(s"POST to $url failed with status $status. Response body: '${response.body}'"))
```
lutnos: Again, not quite sure why this wan't and isn't being dealt with by ErrorHandler

PR #319, opened 2019-02-04
Lg/gfc755/disable hyphen in field id's

LGoldfinch2:
At the top of the FormComponentId case class, the method withSuffix creates a string using a hypen. Does this cause any problems I need to address?


josef-vlach:
This PR is not handling `FormCtx` properly.
Json containing reference to field with underscore fails to parse:

For instance:
```
{
  "id": "anyText",
  "type": "text",
  "label": "Any text",
  "value": "${_textId}"
}
```

PR #330, opened 2019-02-07
Lg/gfc755/disable hyphen in field id's

LGoldfinch2:


app/uk/gov/hmrc/gform/core/parsers/ExprParsers.scala
```diff
@@ -25,7 +25,7 @@ object ExprParsers {
 
   def validateFormCtx(expression: String): Opt[FormCtx] = validateWithParser(expression, expr)
 
-  lazy val expr: Parser[FormCtx] = "${" ~ """\w+""".r ~ "}" ^^ { (loc, _, field, _) =>
+  lazy val expr: Parser[FormCtx] = "${" ~ """[_a-zA-Z][\w]*""".r ~ "}" ^^ { (loc, _, field, _) =>
```
lutnos: You don't need the [] around \w, so [_a-zA-Z]\w* should be sufficient

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormComponentId.scala
```diff
@@ -31,8 +33,17 @@ object FormComponentId {
   implicit val catsEq: Eq[FormComponentId] = Eq.fromUniversalEquals
 
   implicit val vformat: Format[FormComponentId] =
-    ValueClassFormat.vformat("id", FormComponentId.apply, x => JsString(x.value))
+    ValueClassFormat.validatedvformat("id", validate, x => JsString(x.value))
 
   val oformat: OFormat[FormComponentId] = ValueClassFormat.oformat("id", FormComponentId.apply, _.value)
 
+  val idValidationRegex: Regex = """[_a-zA-Z]\w*""".r.unanchored
```
lutnos: This would be better named idValidation, we do not commonly call values "firstNameString", "totalBunsInteger" .

Programmers did once use a convention like that, but it long ago went out of fashion, at least in most languages

lutnos: Also, you don't need the .unanchored here, it is unanchored by default

PR #342, opened 2019-02-13
Add Anonymous validation

PasGat:


test/uk/gov/hmrc/gform/core/TemplateValidatorSpec.scala
```diff
@@ -84,11 +85,11 @@ class TemplateValidatorSpec extends Spec {
           case expr @ HasExpr(SingleExpr(UserCtx(EnrolledIdentifier))) => expr
         }.isEmpty
 
-        whenever(
-          newTemplate.authConfig == HmrcSimpleModule || newTemplate.authConfig
-            .isInstanceOf[HmrcAgentModule] && isAUserCtx) {
-          validateEnrolmentIdentifier(newTemplate) should be(Invalid("Invalid auth type for component type."))
+        whenever(newTemplate.authConfig == HmrcSimpleModule
+          || newTemplate.authConfig.isInstanceOf[HmrcAgentModule] || newTemplate.authConfig == Anonymous && isAUserCtx) {
```
lancewalton: Rather than generating a completely random FormTemplate & then constraining the conditions under which it can apply to the test, replace the parts of the random FormTemplate with what you need. e.g.

```scala
forAll(FormTemplateGen.formTemplateGen, Gen.oneOf(AuthConfigGen.hmrcAgentModuleGen, Gen.const(HmrcSimpleModule)) { (initialTemplate, authConfig) =>
  val template = initialTemplate.copy(authConfig = authConfig)
  ...
```

Not only does this remove the isInstanceOf check, but it also avoids the potential for the spec to fail because the generators tried too many times to generate a FormTemplate that didn't conform to the whenever constraint.

PasGat: makes perfect sense! Done it

app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala
```diff
@@ -155,8 +155,8 @@ object FormTemplateValidator {
     }
 
     formTemplate.authConfig match {
-      case HmrcSimpleModule | HmrcAgentModule(_) if !userContextComponentType.isEmpty =>
-        Invalid("Invalid auth type for component type.")
+      case HmrcSimpleModule | HmrcAgentModule(_) | Anonymous if userContextComponentType.nonEmpty =>
+        Invalid(s"You used ${formTemplate.authConfig} but you didn't provide 'serviceId'.")
```
VlachJosef: This validation should be reversed ie. we should match valid cases, not the invalid ones. Current implementation is wrong wrt to eeitt auth.
It should be something like this:
```
if (userContextComponentType.nonEmpty) {
  formTemplate.authConfig match {
    case HmrcEnrolmentModule(_) | HmrcAgentWithEnrolmentModule(_, _) => Valid
    case _ => Invalid("You used '${user.enrolledIdentifier}' but you didn't provide 'serviceId'.")
  }
} else Valid
```

app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala
```diff
@@ -155,8 +155,8 @@ object FormTemplateValidator {
     }
 
     formTemplate.authConfig match {
-      case HmrcSimpleModule | HmrcAgentModule(_) if !userContextComponentType.isEmpty =>
```
josef-vlach: This validation should be reversed ie. we should match valid cases, not the invalid ones. Current implementation is wrong wrt to eeitt auth.
It should be something like this:
```
if (userContextComponentType.nonEmpty) {
  formTemplate.authConfig match {
    case HmrcEnrolmentModule(_) | HmrcAgentWithEnrolmentModule(_, _) => Valid
    case _ => Invalid("You used '${user.enrolledIdentifier}' but you didn't provide 'serviceId'.")
  }
} else Valid
```This validation should be reversed ie. we should match valid cases, not the invalid ones. Current implementation is wrong wrt to eeitt auth.
It should be something like this:
```
if (userContextComponentType.nonEmpty) {
  formTemplate.authConfig match {
    case HmrcEnrolmentModule(_) | HmrcAgentWithEnrolmentModule(_, _) => Valid
    case _ => Invalid("You used '${user.enrolledIdentifier}' but you didn't provide 'serviceId'.")
  }
} else Valid
```

Comment (PasGat) [reply app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala]: What about 'Anonymous'?
should not be 
"case HmrcSimpleModule | HmrcAgentModule(_) | Anonymous => "

Comment (josef-vlach) [reply app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala]: Combination of `${user.enrolledIdentifier}` and `anonymous` is indeed invalid as per my snippet.

Comment (PasGat) [reply app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala]: I have changed the implementation + . I gave a good refactor to the test - however there is a test for Anonymous scenario missing. Anyway let me know

File: app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala [hunk 4]
```diff
@@ -149,17 +149,19 @@ object FormTemplateValidator {
...
     }
 
-    formTemplate.authConfig match {
-      case HmrcSimpleModule | HmrcAgentModule(_) if !userContextComponentType.isEmpty =>
-        Invalid("Invalid auth type for component type.")
-      case _ => Valid
-    }
-  }
+  def validateEnrolmentIdentifier(formTemplate: FormTemplate): ValidationResult =
+    if (userContextComponentType(formTemplate.expandFormTemplate.allFCs).nonEmpty) {
```
josef-vlach: More clear way would be to use `formTemplate.expandFormTemplate.allFCs.exists(...)`. There is no need to traverse whole list.

PR #346, opened 2019-02-14
Add test for Anonymous authConf + refactor test

PasGat:


app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala
```diff
@@ -149,14 +149,14 @@ object FormTemplateValidator {
     }
   }
 
-  val userContextComponentType: List[FormComponent] => List[FormComponent] =
+  val userContextComponentType: List[FormComponent] => Boolean =
     enrolledIdentifierComponents =>
       enrolledIdentifierComponents.collect {
```
josef-vlach: More clear way would be to use `enrolledIdentifierComponents.exists(...)`. There is no need to filter/map whole list and asking if result list is empty.

test/uk/gov/hmrc/gform/core/TemplateValidatorSpec.scala
```diff
@@ -24,9 +24,13 @@ import org.scalacheck.Gen
...
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.destinations.Destinations
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.generators._
-import uk.gov.hmrc.gform.sharedmodel.formtemplate.generators.FormComponentGen._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.generators.PrimitiveGen._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.generators.SectionGen._
+import FormTemplateValidator._
+import AuthConfigGen._
+import FormComponentGen._
+
+import scala.collection.immutable.List
```
josef-vlach: Is this import needed? Isn't it included in Predef?

test/uk/gov/hmrc/gform/core/TemplateValidatorSpec.scala
```diff
@@ -69,31 +73,29 @@ class TemplateValidatorSpec extends Spec {
...
 
-      whenever(isAUserCtx.isEmpty) {
-        validateEnrolmentIdentifier(newTemplate) should be(Valid)
+        whenever(!isAUserCtx) {
+          validateEnrolmentIdentifier(newTemplate, userContextComponentType) should be(Valid)
+        }
+    }
+  }
+
+  it should "invalid with Anonymous" in {
```
josef-vlach: If we want to test Anonymous (ie. `anonymousGen`), we should check also for other cases of `AuthConfig`.

PR #348, opened 2019-02-15
GFC-783/removed submit from submissioncontroller aswell as the routes…

DarylMConway:
GFC-783/removed submissionWithoutPDF as it was no longer used


DarylMConway:
removed the pdf generation that was never being called. putting branch on hold and focusing on updating systems to play 2.5 moving ticket back to "To Do" as requested.

PR #353, opened 2019-02-19
GFC-767, GFC-768: Allow composite keys in lookups.

lancewalton:


app/uk/gov/hmrc/gform/submission/handlebars/LookupMatchStringParser.scala
```diff
@@ -0,0 +1,69 @@
...
+      case _ ~ keyValues ~ _ => keyValues
+    }
+
+    def quotedStringParser: Parser[String] = "'" ~ charSequenceParser ~ "'" ^^ {
+      case _ ~ charSequence ~ _ => charSequence
+    }
+
+    def caseParser: Parser[Option[String]] = compositeKeyParser ~ "=>" ~ quotedStringParser ~ ";?".r ^^ {
+      case caseKey ~ _ ~ resultValue ~ _ =>
+        if (isMatchingCase(caseKey)) Option(resultValue)
```
VlachJosef: Using `Option.apply` is throwing away information about source of possible `None` value.  Written like this one needs to assume that `resultValue` may be possible be `null`. In this case I think we are save to use `Some(resultValue)` and spare reader to wonder why we handle possible `null`s.

lancewalton: Sorted

app/uk/gov/hmrc/gform/submission/handlebars/HandlebarsTemplateProcessorHelpers.scala
```diff
@@ -361,20 +361,11 @@ class HandlebarsTemplateProcessorHelpers(timeProvider: TimeProvider = new TimePr
...
-    condition(
-      trimAll(options.split(','))
-        .map(parseBranch)
-        .foldLeft(Map.empty[String, String]) { _ ++ _ }
-        .getOrElse(value, null))
+  def lookup(o: Options): CharSequence = {
+    val matchString = o.params(0).toString
+    val key: List[String] = o.params.drop(1).toList.collect { case s: String => s }
+    LookupMatchStringParser(matchString, key).getOrElse(
+      throw new Exception(s"Attempt to lookup $key failed in $matchString"))
```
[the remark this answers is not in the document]
lancewalton: This method is the entry point for the handlebars processor when it calls the helper function. We can't change the return type (because all handlebars helper functions are required to return CharSequence), & we do want to signal strongly that an error has occurred.

PR #354, opened 2019-02-19
Added build.sbt file and removed MicroService and MicroServiceBuild

Davidrymer97:


build.sbt
```diff
@@ -0,0 +1,55 @@
+import Dependencies.appDependencies
+import TestPhases.oneForkedJvmPerTest2
+import Resolvers._
+import uk.gov.hmrc.DefaultBuildSettings.{ addTestReportOption, defaultSettings, scalaSettings }
+import uk.gov.hmrc.sbtdistributables.SbtDistributablesPlugin.publishingSettings
+
```
arkowalski: do we need scoverageSettigs for our coverage to take effect?
e.g.
lazy val scoverageSettings = {
  import scoverage.ScoverageKeys
  Seq(
    // Semicolon-separated list of regexs matching classes to exclude
    ScoverageKeys.coverageExcludedPackages := """uk\.gov\.hmrc\.BuildInfo;.*\.Routes;.*\.RoutesPrefix;.*Filters?;MicroserviceAuditConnector;Module;GraphiteStartUp;.*\.Reverse[^.]*""",
    ScoverageKeys.coverageMinimum := 80.00,
    ScoverageKeys.coverageFailOnMinimum := true,
    ScoverageKeys.coverageHighlighting := true,
    parallelExecution in Test := false
  )
}

build.sbt
```diff
@@ -0,0 +1,55 @@
...
+lazy val playSettings: Seq[Setting[_]] = Seq.empty
+
+name := "gform"
```
arkowalski: also add "    organization := "uk.gov.hmrc"

project/Resolvers.scala
```diff
@@ -0,0 +1,9 @@
+import sbt._
+
+object Resolvers {
```
arkowalski: remove this, just put the content in build.sbt

project/TestPhases.scala
```diff
@@ -0,0 +1,10 @@
+import sbt.{ForkOptions, TestDefinition}
+import sbt.Tests.{Group, SubProcess}
+
+object TestPhases {
+
+  def oneForkedJvmPerTest2(tests: Seq[TestDefinition]) =
```
arkowalski: why is there a "2" in the name ?

arkowalski: also currently there is only one test phase, move this function into build.sbt and make it private

lutnos: At one time, at least, there was an existing oneForkedJvmPerTest which we could not override.  If it works, then there is no reason why it should not be oneForkedJvmPerTest now

PR #359, opened 2019-02-25
Kyle/gfc 791/cache tax period

kylemchownhmrc:


app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -49,3 +50,41 @@ case class TaxResponse(id: HmrcTaxPeriod, obligation: Obligation)
...
   implicit val format: OFormat[TaxResponse] = Json.format[TaxResponse]
 }
+
+case class TaxPeriodIdentifier(idType: IdType, idNumber: IdNumber, regimeType: RegimeType)
+
+object TaxPeriodIdentifier {
+  implicit val format: OFormat[TaxPeriodIdentifier] = Json.format[TaxPeriodIdentifier]
+}
+
+case class AllInfo(
```
lutnos: This name needs to be better.  Perhaps something along the lines of TaxPeriodMapping

app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
```diff
@@ -41,7 +41,8 @@ case class Form(
   formData: FormData,
   status: FormStatus,
   visitsIndex: VisitIndex,
-  envelopeExpiryDate: Option[EnvelopeExpiryDate]
+  envelopeExpiryDate: Option[EnvelopeExpiryDate],
+  obligations: Option[ListAllInfo]
```
lutnos: Is there a meaningful difference between None and Some(List.empty) ?  If not, then it should either not be an Option, or should be an Option of a NonEmptyList

lutnos: In fact, this ought not be either a List[], or an Option[NonEmptyList], it should be

obligations: NonEmptyList[TaxPeriodInformation]

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -49,3 +50,28 @@ case class TaxResponse(id: HmrcTaxPeriod, obligation: Obligation)
...
+  periodKey: String)
+
+object TaxPeriodInformation {
+  implicit val format: OFormat[TaxPeriodInformation] = derived.oformat
+}
+
+sealed trait Obligations
+final case object NotChecked extends Obligations
+final case object NoObligations extends Obligations
+final case class RetrievedObligations(listOfObligations: List[TaxPeriodInformation]) extends Obligations
```
lutnos: That needs to be a NonEmptyList[TaxPeriodInformation]]

app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
```diff
@@ -322,7 +322,7 @@ class FormComponentMaker(json: JsValue) {
 
   private lazy val hmrcTaxPeriodOpt: Opt[HmrcTaxPeriod] = {
     val oHmrcTaxPeriod: Opt[HmrcTaxPeriod] = (idType, idNumber, regimeType) match {
-      case (Some(a), Some(b), Some(c)) => HmrcTaxPeriod(IdType(a), IdNumber(b), RegimeType(c)).asRight
+      case (Some(a), Right(Some(b: TextExpression)), Some(c)) => HmrcTaxPeriod(IdType(a), b, RegimeType(c)).asRight
```
josef-vlach: What about all other cases? We are not handling them. I've never saw this from the compiler:

```
[warn]  [E3] app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala:324:46:
[warn]       Exhaustivity analysis reached max recursion depth, not all missing cases are reported.
[warn]       (Please try with scalac -Ypatmat-exhaust-depth 40 or -Ypatmat-exhaust-depth off.)
[warn]           val oHmrcTaxPeriod: Opt[HmrcTaxPeriod] = (idType, idNumber, regimeType) match {
```

josef-vlach:
@lutnos @kylemchownhmrc How come this got merged? I don't think solution in [`6d8939d`](/hmrc/gform/commit/6d8939d9ea07dba2da987783c23df21694e1a91b) for my comment is correct, quite the contrary actually it make things even worse.

lutnos:
@josef-vlach 

@kylemchownhmrc and I discussed recalculation here & we now believe we understand the problem and a proper solution for hmrcTaxPeriod .  We have it on a whiteboard and can share it with you tomorrow.  It will need a more that trial change.

For now, I have merged, although there may be a remaining issue with an empty NEL & passing an Expression to the backend is unsatisfactory.

I have done this because the current code will satisfy the AC for GFC-791, that an agent can enter a registration number and get an HmrcTaxPeriod.  This will satisfy our immediate requirement to have an automated Aggregate Levy return working.

As I say, we would like to discuss with you tomorrow

PR #366, opened 2019-02-26
GFC-631: Asynchronous ROSM enrolment destination connector

lancewalton:


app/uk/gov/hmrc/gform/submission/handlebars/MdtpHttpClient.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package uk.gov.hmrc.gform.submission.handlebars
+
+import uk.gov.hmrc.gform.wshttp.JsonHttpClient
+
+case class MdtpHttpClient[F[_]](clients: Map[MdtpServiceName, JsonHttpClient[F]]) {
+  def select(serviceName: MdtpServiceName): JsonHttpClient[F] = clients(serviceName)
+}
+
+object MdtpHttpClient
```
josef-vlach: Is there any reason why this object exists?

lancewalton: Initially it was wrapping an Either[HttpClient, Map[MdtpServiceName, HttpClient]], which seemed to me to need some semantics to be provided by wrapping it. However, it got simplified, so maybe it's not needed now. Shall I remove it?

josef-vlach: I was thinking it may exists to override default companion object for some scala magic reasons. If it is not needed than I'd say it's better to remove it.

lancewalton: OK

PR #369, opened 2019-02-27
Gfc/783 without pdf

DarylMConway:


app/uk/gov/hmrc/gform/submission/SubmissionService.scala
```diff
@@ -83,11 +73,6 @@ class SubmissionService(
       )
     )
 
-  private def getSignedForm(formId: FormId)(implicit hc: HeaderCarrier) = formService.get(formId).flatMap {
-    case f @ Form(_, _, _, _, _, Signed, _, _, _) => Future.successful(f)
-    case _                                        => Future.failed(new Exception(s"Form $FormId status is not signed"))
-  }
-
   private def submission(
```
josef-vlach: Existence of this method is no longer justified, it can (and it should) be inlined into `submissionWithPdf` method.

DarylMConway: getSignedForm and submissionWithPdf are both calling formService but getSignedForm is flat mapping it to check if it is the correct format.
why is it needed in submissionWithPdf now as it was never used in was it prior to this?

PR #376, opened 2019-02-28
Allows you to have multiple components in the same form

kylemchownhmrc:


app/uk/gov/hmrc/gform/form/FormService.scala
```diff
@@ -64,10 +63,7 @@ class FormService(save4Later: Save4Later) {
           status = newStatus(form, userData.formStatus),
           visitsIndex = userData.visitsIndex,
           obligations = userData.obligations)
-      _ <- {
-        Logger.debug(Json.prettyPrint(Json.toJson(newForm)) + "UpdateUserData")
```
arkowalski: was this meant to be removed?

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -70,9 +74,32 @@ object TaxPeriodInformation {
...
 final case object NotChecked extends Obligations
-final case class RetrievedObligations(listOfObligations: NonEmptyList[TaxPeriodInformation]) extends Obligations
+final case class RetrievedObligations(listOfObligations: List[TaxPeriodInformation]) extends Obligations
 
 object Obligations {
-  import JsonUtils._
   implicit val format: OFormat[Obligations] = derived.oformat
 }
+
+case class IdNumberValue(value: String) extends AnyVal
```
arkowalski: do we need "extends AnyVal" maybe we can avoid it.

app/uk/gov/hmrc/gform/form/FormService.scala
kylemchownhmrc: Yes. It was added as a way to check what was actually reaching the backend when a call was made. After the issue was resolved it was removed

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
kylemchownhmrc: As far as I know when we have a case class that takes in only a single basic parameter we are meant to extend AnyVal. I'm not too sure why so if you have a reason to remove it we can

PR #388, opened 2019-03-06
GFC-819: toDesDate handlebars helper function

lancewalton:


app/uk/gov/hmrc/gform/submission/handlebars/HandlebarsTemplateProcessorHelpers.scala
```diff
@@ -44,12 +44,23 @@ class HandlebarsTemplateProcessorHelpers(timeProvider: TimeProvider = new TimePr
...
-    (for {
+    extractDateFields(dateFieldId, o) { (day, month, year) =>
+      s"$year${padLeft(month, 2, '0')}${padLeft(day, 2, '0')}"
+    }.map(condition) getOrElse NullString
+  }
+
+  def toDesDate(dateFieldId: String, o: Options): CharSequence = log("toDesDate", dateFieldId) {
+    extractDateFields(dateFieldId, o) { (day, month, year) =>
+      s"$year-${padLeft(month, 2, '0')}-${padLeft(day, 2, '0')}"
+    }.map(condition) getOrElse NullString
```
arkowalski: using nulls sounds like a bad idea? should we throw an exception maybe? if we cannot convert to the type? 
I dont think this getOrElse should ideally never return the else NullString. for both EtmpDate and DesDate.

lancewalton: It's not actually a null. It's a way of carrying nullness as a CharSequence through the Handlebars machinery. The result of this function is embedded in the template in place of the Handlebars stuff that invoked it. We can't know what purpose the caller has, so we can't make a decision about what to do about it. For example, they may do {{isNull (toDesDate 'myDate')}}, to gracefully handle the null.

I agree this all sucks quite a bit.

arkowalski: I see you're not the one that made that possible so ill approve, but would rather not see nulls

lancewalton: Yes. The problem is that Handlebars was originally a JavaScript library where null occupies a space somewhere between 'not a value' & 'the ill defined identity value for any particular type you might care for'. This idea has been carried into the Java Handlebars implementation, but the mismatch between the type systems necessitates some gymnastics to try to keep things working.

PR #389, opened 2019-03-06
GFC-696 - EEITT Reg/Subscriptions - Replace UTR/Postcode validator wi…

josef-vlach:
…th hmrcRosmRegistrationCheck

### DIFF
--- a/app/uk/gov/hmrc/gform/des/DesConnector.scala
+++ b/app/uk/gov/hmrc/gform/des/DesConnector.scala
@@ -16,13 +16,19 @@
 
 package uk.gov.hmrc.gform.des
 
+import cats.syntax.eq._
+import cats.instances.string._
+import cats.instances.int._
 import play.api.Logger
 import play.api.http.Status
 import play.api.libs.json._
 import scala.concurrent.{ ExecutionContext, Future }
+import scala.reflect.runtime.universe.TypeTag
+import scala.util.{ Failure, Success, Try }
 import uk.gov.hmrc.gform.auditing.loggingHelpers
 import uk.gov.hmrc.gform.config.DesConnectorConfig
-import uk.gov.hmrc.gform.sharedmodel.des.{ DesRegistrationRequest, DesRegistrationResponse }
+import uk.gov.hmrc.gform.sharedmodel.{ NotFound, ServiceCallResponse, ServiceNotAvailable, ServiceResponse }
+import uk.gov.hmrc.gform.sharedmodel.des.{ DesRegistrationRequest, DesRegistrationResponse, DesRegistrationResponseError }
 import uk.gov.hmrc.gform.wshttp.WSHttp
 import uk.gov.hmrc.gform.sharedmodel.Obligation
 import uk.gov.hmrc.http._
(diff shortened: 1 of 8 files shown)


app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -31,17 +37,56 @@ import uk.gov.hmrc.http.logging.Authorization
...
-      desRegistrationRequest)
 
+    wSHttp
+      .doPost[DesRegistrationRequest](
+        s"$baseUrl${desConfig.basePath}/registration/organisation/utr/$utr",
+        desRegistrationRequest,
+        List.empty[(String, String)]
+      )
+      .map { httpResponse =>
+        if (httpResponse.status === 400) {
```
arkowalski: this potentially will be an Upstream4xxResponse that you would have to recover on.

app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -31,17 +37,56 @@ import uk.gov.hmrc.http.logging.Authorization
...
+          processResponse[DesRegistrationResponseError, Nothing](httpResponse) { desError =>
+            if (desError.code === "INVALID_UTR") NotFound
+            else {
+              Logger.error("Problem when calling des registration: " + Json.prettyPrint(Json.toJson(desError)))
```
arkowalski: not sure what desError returns are there any concerns there around sensitive data?

app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -31,17 +37,56 @@ import uk.gov.hmrc.http.logging.Authorization
...
+              ServiceNotAvailable
```
arkowalski: this is still within the scope of if (httpResponse.status === 400) 
I don't imagine responseStatus 400 representing service unavailable

app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -31,17 +37,56 @@ import uk.gov.hmrc.http.logging.Authorization
...
+          Logger.error(s"Unknown problem when calling des registration. Expected json for $tpe, but got: $jsError")
+          ServiceNotAvailable
+        }
+      case Failure(failure) =>
+        val tpe = implicitly[TypeTag[A]].tpe
+        val status = httpResponse.status
+        Logger.error(
+          s"Unknown problem when calling des registration. Response status was: $status. Expected json for $tpe, but got :",
+          failure)
+        ServiceNotAvailable
```
arkowalski: bad json does not mean ServiceNotAvailable, seems that this could give some confusion

app/uk/gov/hmrc/gform/des/DesConnector.scala
```diff
@@ -31,17 +37,56 @@ import uk.gov.hmrc.http.logging.Authorization
 class DesConnector(wSHttp: WSHttp, baseUrl: String, desConfig: DesConnectorConfig) {
 
   def lookupRegistration(utr: String, desRegistrationRequest: DesRegistrationRequest)(
-    implicit ex: ExecutionContext): Future[DesRegistrationResponse] = {
+    implicit ex: ExecutionContext): Future[ServiceCallResponse[DesRegistrationResponse]] = {
```
arkowalski: you can just call the class "ServiceResponse"

app/uk/gov/hmrc/gform/sharedmodel/ServiceCallResponse.scala
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.gform.sharedmodel
+
+import play.api.libs.json._
+
+sealed trait ServiceCallResponse[+A] extends Product with Serializable
```
arkowalski: why do we even need this? seems like this could be avoided? if its all fine just return 200? and for others appropriate code?

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
```diff
@@ -27,9 +27,9 @@ sealed trait Validator {
 case object Validator {
   private val templateReads: Reads[Validator] = Reads { json =>
     (json \ "validatorName").as[String] match {
-      case "hmrcUTRPostcodeCheck"    => json.validate[HMRCUTRPostcodeCheckValidator]
-      case "bankAccountModulusCheck" => json.validate[BankAccoutnModulusCheck]
-      case unsupported               => JsError("Unsupported '" + unsupported + "' kind of validator.")
+      case "hmrcRosmRegistrationCheck" => json.validate[HMRCUTRPostcodeCheckValidator]
```
arkowalski: rename HMRCUTRPostcodeCheckValidator to something more uniform now

app/uk/gov/hmrc/gform/des/DesConnector.scala
josef-vlach: I agree with what you are saying, on the other hand bad json doesn't mean utr + postcode combination is invalid. As far as user is concerned we need to handle two cases.

1. utr + postcode combination is valid - ie we get response which we can turn into `DesRegistrationResponse`
2. utr + poscode combination is invalid - ie status=400 and we get response which we can turn into `DesRegistrationResponseError` and on top of that `response.code === "INVALID_UTR"`
3. any other case (being it wrong json, service being down etc.) we want to treat as technical difficulties (from user perspective).

To represent case 3. I'm using `ServiceNotAvailable` object, do you think renaming this object to `InvalidResponse` would be more appropriate? Any other ideas for name of that object?

app/uk/gov/hmrc/gform/des/DesConnector.scala
josef-vlach: No, there are no concerns about sensitive data.

app/uk/gov/hmrc/gform/des/DesConnector.scala
josef-vlach: `Upstream4xxResponse` cannot be used since we need to treat `code === "INVALID_UTR"` in different way from any other `code` value. All of them has `status === 400`

app/uk/gov/hmrc/gform/des/DesConnector.scala
josef-vlach: `ServiceResponse` is data constructor for `ServiceCallResponse` data type. (here I'm using haskell vocabulary)

app/uk/gov/hmrc/gform/des/DesConnector.scala
josef-vlach: Please see https://github.com/hmrc/gform/pull/389/files/6d17606692afce5f72fbbc89d85f3c3683449b9a#r263312198

app/uk/gov/hmrc/gform/sharedmodel/ServiceCallResponse.scala
josef-vlach: This code exists to lift http protocol intricacies into type level. 

In a long run I'd like to see migration of all our calls to third party apis (ie. not platform calls or destinations related) to this data type.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Validator.scala
josef-vlach: will do

app/uk/gov/hmrc/gform/des/DesConnector.scala
arkowalski: yes this is a bit confusing because of that, you can handle this on the frontend and return ServiceNotAvailable on that service but the backend should reflect what is actually happening as the user is not seeing that, and its easier for developers to diagnose

PR #390, opened 2019-03-08
Kyle/gfc 791/multiple components in one form

kylemchownhmrc:


app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
```diff
@@ -51,8 +53,16 @@ object Form {
...
 
+  val readerRetreived: Reads[Option[Obligations]] = (__ \ "RetrievedObligations" \ "listOfObligations")
+    .readNullable[List[TaxPeriodInformation]]
+    .map(_.map(RetrievedObligations))
+  val readerStandard: Reads[Option[Obligations]] =
+    (__ \ "obligations").readNullable[List[TaxPeriodInformation]].map(_.map(RetrievedObligations))
   val readObligations: Reads[Obligations] =
-    (__ \ "obligations").readNullable[Obligations].map(a => a.getOrElse(NotChecked))
+    readerRetreived.orElse(readerStandard).map { a =>
+      println("AAAA" + a)
```
arkowalski: clean this

app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
```diff
@@ -20,9 +20,11 @@ import julienrf.json.derived
 import play.api.libs.json.Reads._
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
-import uk.gov.hmrc.gform.sharedmodel.{ NotChecked, Obligations, UserId }
+import uk.gov.hmrc.gform.sharedmodel._
```
arkowalski: try avoiding "_" if possible

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -20,9 +20,12 @@ import java.util.Date
 
 import cats.data.NonEmptyList
 import julienrf.json.derived
+import play.api.data.validation.ValidationError
 import play.api.libs.json._
 import uk.gov.hmrc.gform.sharedmodel.formtemplate._
 
+import scala.collection.immutable.{ ::, List, Nil }
```
josef-vlach: No need for this

app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
```diff
@@ -20,9 +20,11 @@ import julienrf.json.derived
...
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ FormTemplateId, SectionNumber }
 
+import scala.collection.immutable.List
```
josef-vlach: No need for this

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -70,9 +74,32 @@ object TaxPeriodInformation {
...
+
+  implicit def readsNonEmptyList[T: Reads] = Reads[NonEmptyList[T]] { json =>
+    Json.fromJson[List[T]](json).flatMap {
+      case Nil     => JsError(ValidationError(s"Required at least one element. Got: $json"))
+      case x :: xs => JsSuccess(NonEmptyList(x, xs))
+    }
+  }
+
+  implicit def writesNonEmptyList[T: Writes] = Writes[NonEmptyList[T]] { v =>
+    JsArray((v.head :: v.tail).map(Json.toJson(_)).toList)
```
arkowalski: v.head and v.tail  can we get a NullPointerException to these if v.size == 0 ?

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
```diff
@@ -70,9 +74,32 @@ object TaxPeriodInformation {
...
+
+object IdNumberValue {
+  implicit val format: OFormat[IdNumberValue] = derived.oformat
+}
+
+case class HmrcTaxPeriodWithEvaluatedId(hmrcTaxPeriod: HmrcTaxPeriod, idNumberValue: IdNumberValue)
+
+object HmrcTaxPeriodWithEvaluatedId {
+
+  implicit def readsNonEmptyList[T: Reads] = Reads[NonEmptyList[T]] { json =>
```
josef-vlach: JsonUtils provide `Reads[NonEmptyList[T]]`, use that instead.

app/uk/gov/hmrc/gform/sharedmodel/TaxPeriods.scala
kylemchownhmrc: Not sure. This is for writing a NonEmptyList though so size should never be 0.
Also, this code is almost directly copied from a different reads and writes so I imagine if it's fine there it should also be here

app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
```diff
@@ -51,8 +53,15 @@ object Form {
   val readVisitIndex: Reads[VisitIndex] =
     (__ \ "visitsIndex").readNullable[List[Int]].map(a => VisitIndex(a.fold(Set.empty[Int])(_.toSet)))
 
+  val readerRetreived: Reads[Option[Obligations]] = (__ \ "RetrievedObligations" \ "listOfObligations")
+    .readNullable[List[TaxPeriodInformation]]
+    .map(_.map(RetrievedObligations))
+  val readerStandard: Reads[Option[Obligations]] =
+    (__ \ "obligations").readNullable[List[TaxPeriodInformation]].map(_.map(RetrievedObligations))
```
josef-vlach: There is something fishy going on. Why we read `Obligations` at one place as `"RetrievedObligations" \ "listOfObligations"` and at second place `"obligations"`?

arkowalski: this is what I was asking you about before lunch, if we do not have the reads \ "obligations" the frontend gets an error page thrown..

arkowalski: I imagine this would be something related with how the data is stored and then attempted to retrieve.

test/uk/gov/hmrc/gform/sharedmodel/form/SubmissionSpec.scala
```diff
@@ -18,7 +18,7 @@ package uk.gov.hmrc.gform.sharedmodel.form
 
 import play.api.libs.json._
 import uk.gov.hmrc.gform.Spec
-import uk.gov.hmrc.gform.sharedmodel.ExampleData
+import uk.gov.hmrc.gform.sharedmodel.{ExampleData, RetrievedObligations, TaxPeriodInformation}
```
lutnos: No needed

app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
josef-vlach: Yes, of course, data in master are using `"obligations"` as a key in json. We need to write unit test to cover if we are supporting all possible states.

1. Having no obligation in mongo.
2. Having obligations under `"obligations"` key (as result of using `Json.format` macro).
3. Having obligations under `"RetrievedObligations" \ "listOfObligations"` key (as result of using  `derived.oformat` macro)

PR #418, opened 2019-04-12
added legacyFcEnrolmentVerifier

Davidrymer97:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/AuthConfig.scala
```diff
@@ -29,7 +29,8 @@ object EEITTAuthConfig {
 
 case class EnrolmentAuth(
   serviceId: ServiceId,
-  enrolmentCheck: EnrolmentCheck
+  enrolmentCheck: EnrolmentCheck,
+  legacyFcEnrolmentVerifier: Option[LegacyFcEnrolmentVerifier]
```
josef-vlach: `LegacyFcEnrolmentVerifier` should not be part of `EnrolmentAuth`. Spec says that `LegacyFcEnrolmentVerifier` is valid only when Enrolment section exists. 

Existence of Enrolment section is witnessed by instance of `RequireEnrolment`, so we need to put `LegacyFcEnrolmentVerifier` into `RequireEnrolment`. This is what I understand as TDD (Type Driven Development).

josef-vlach: By using `Option` for `legacyFcEnrolmentVerifier` we are losing readability in a sense that we are using `None` as its value here and there and from a look at such a code it is hard to guess what type of `None` really is.

We can increase readability by introducing specific type which will be isomorphic with `Option[LegacyFcEnrolmentVerifier]` to alleviate this problem. 

We've used this same technique to have better representation of `Boolean` here [ObligationsAction.scala](https://github.com/hmrc/gform-frontend/blob/master/app/uk/gov/hmrc/gform/models/gform/ObligationsAction.scala#L19-L21)

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/AuthConfig.scala
```diff
@@ -134,20 +137,32 @@ object AuthConfig {
   def toEnrolmentPostCheck(maybeRegimeId: Option[RegimeId]): EnrolmentPostCheck =
     maybeRegimeId.fold(NoCheck: EnrolmentPostCheck)(RegimeIdCheck.apply)
 
+  def enrolmentActionMatch(enrolmentAction: Option[EnrolmentAction]): EnrolmentAction = enrolmentAction match {
```
josef-vlach: The same as on frontend: 

Body of this method may be shortened to just `enrolmentAction.getOrElse(NoAction)`

PR #425, opened 2019-04-25
GFC-713 RevealingChoice component

lancewalton:
I picked this work up from Sean, so other people generally looking over it would be good.

I'm also looking for help. I'm concerned that everywhere there's a pattern match for FormComponent.`type` that treats Group as a special case may also need RevealingChoice to be treated as a special case.


josef-vlach:
@lancewalton @PasGat Can you review this please? Thanks in advance.

PR #427, opened 2019-04-26
[GFC-878][DC] added shortText(min,max)

DarylMConway:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormatExpr.scala
```diff
@@ -150,7 +150,10 @@ final case class PositiveNumber(
     extends TextConstraint
 
 case object BasicText extends TextConstraint
-case object ShortText extends TextConstraint
+case class ShortText(
+  min: Int = TextConstraint.defaultShortTextMinChars,
+  max: Int = TextConstraint.defaultShortTextMaxChars)
```
josef-vlach: Can we remove default values from here, we do not need them.

app/uk/gov/hmrc/gform/formtemplate/FormComponentMakerService.scala
```diff
@@ -47,9 +47,9 @@ object FormComponentMakerService {
     toUpperCase: UpperCaseBoolean) = (maybeFormatExpr, maybeValueExpr, displayWidth) match {
     // format: off
     case (Some(TextFormat(f)), HasTextExpression(expr), None)                => Text(f, expr, DisplayWidth.DEFAULT, toUpperCase).asRight
-    case (None,                HasTextExpression(expr), None)                => Text(ShortText, expr, DisplayWidth.DEFAULT, toUpperCase).asRight
+    case (None,                HasTextExpression(expr), None)                => Text(ShortText(), expr, DisplayWidth.DEFAULT, toUpperCase).asRight
```
josef-vlach: Let's not rely on default values like in `ShortText()` at all. Let's introduce something like `ShortText.default` instead.

```
object ShortText {
  val default = ShortText(0, 1000)
}
```

Ask @Davidrymer97 how the other day we've spent whole morning hunting for a bug which turned out to be caused by default value in a case class.

PR #437, opened 2019-06-03
GFC-892: add Seed type to generate submission references

co-category:


app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -81,7 +81,7 @@ class ApplicationModule(context: Context) extends BuiltInComponentsFromContext(c
   lazy val save4later =
     new Save4Later(shortLivedCacheModule.shortLivedCache)
 
-  lazy val formService = new FormService(save4later, fileUploadModule.fileUploadService)
+  lazy val formService = new FormService(save4later, fileUploadModule.fileUploadService, configModule)
```
lancewalton: I believe this is the first instance of us injection the configModule into a *service*, rather than given the service what it needs from the configModule. Do we want to go this way?

app/uk/gov/hmrc/gform/sharedmodel/form/Seed.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package uk.gov.hmrc.gform.sharedmodel.form
+
+import java.util.UUID
+
+import play.api.libs.json._
+import uk.gov.hmrc.gform.sharedmodel.ValueClassFormat
+
+case class Seed(value: String)
```
lancewalton: I don't understand what this class is giving us. Why don't we just instantiate the Form a SubmissionReference instance, rather than giving it this Seed thing & then deriving a SubmissionReference from it?

If this has been discussed & everybody is happy with it, feel free to ignore this comment.

app/uk/gov/hmrc/gform/form/FormService.scala
```diff
@@ -47,11 +51,11 @@ class FormService[F[_]: Monad](formPersistence: FormPersistenceAlgebra[F], fileU
     val formId = FormId(userId, formTemplateId, accessCode)
     val expiryDate = timeProvider.localDateTime().plusDays(expiryDays)
 
-    for {
-      envelopeId <- fileUpload.createEnvelope(formTemplateId, expiryDate)
-      form = Form(
+    if (config.env == "OFSTED") {
```
josef-vlach: This `if ... else` should be completely removed and creation of `envelopeId` should be put as a dependency instead. In this way `FormService` may stay oblivious to what environment it runs in.

PR #478, opened 2019-09-17
GFC-1081: Includeifs to work with multivalue

lancewalton:
I'm not sure of this implementation.

In particular, the first parameter of BooleanExpr/Includes should be a FormCtx not an Expression, but I can't get the BooleanExprParser to play nicely.

Unrelated: We should probably move the majority of the contents of the BooleanExpr.scala to somewhere outside of sharedmodel.


app/uk/gov/hmrc/gform/core/parsers/BooleanExprParser.scala
```diff
@@ -60,6 +60,9 @@ object BooleanExprParser {
     | exprFormCtx ~ ">" ~ exprFormCtx ^^ { (loc, expr1, op, expr2) =>
       GreaterThan(expr1, expr2)
     }
+    | exprFormCtx ~ "includes" ~ exprFormCtx ^^ { (_, expr1, _, expr2) =>
```
josef-vlach: We can change this to
```
| alphabeticOnly ~ "includes" ~ exprFormCtx ^^ { (_, expr1, _, expr2) =>
      Includes(FormCtx(expr1), expr2)
```
where logic later will verify that whatever is captured by `alphabeticOnly` is valid `FormComponentId`.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/BooleanExpr.scala
```diff
@@ -35,6 +35,7 @@ final case class Or(left: BooleanExpr, right: BooleanExpr) extends BooleanExpr
 final case class And(left: BooleanExpr, right: BooleanExpr) extends BooleanExpr
 final case object IsTrue extends BooleanExpr
 final case object IsFalse extends BooleanExpr
+final case class Includes(multiValueField: Expr, value: Expr) extends BooleanExpr
```
josef-vlach: We should make `multiValueField` of type `FormCtx` (as you say) to reject invalid expressions. See previous comment as to how.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/BooleanExpr.scala
```diff
@@ -75,6 +76,11 @@ object BooleanExpr {
         case _ => ("", List.empty)
       }
 
+    def multiValue(collectionFieldId: Expr): (List[String], List[FormComponentId]) = {
+      val (collection, fieldId) = stringValue(collectionFieldId)
```
josef-vlach: There is this `data: Map[FormComponentId, Seq[String]]` used in `stringValue`, and I think this ticket is great opportunity to amend previous errors of not embracing that type.

What I mean by this is that value of `Seq[String]` type should be holding all selections like `List("1", "2", "3")`, and not concatenated string of selections like List("1,2,3") as it currently does.

PR #488, opened 2019-11-22
GFORMS-114: Upgrade Play to 2.6

lancewalton:


conf/application.conf
```diff
@@ -116,7 +116,7 @@ auditing {
     enabled=false
     traceRequests=true
```
josef-vlach: Can we remove `traceRequests=true`, it is not used.

PR #491, opened 2019-12-09
GFORMS-346 - DMS submission is failing for "not visible" files

josef-vlach:


app/uk/gov/hmrc/gform/fileupload/FileUploadService.scala
```diff
@@ -61,7 +58,7 @@ class FileUploadService(
     val fileNamePrefix = s"${submissionRef.withoutHyphens}-$date"
     val reconciliationId = ReconciliationId.create(submissionRef)
     val metadataXml = MetadataXml.xmlDec + "\n" + MetadataXml
-      .getXml(submission, reconciliationId, summaries.pdfSummary, dmsSubmission, numberOfAttachments)
+      .getXml(submission, reconciliationId, summaries.pdfSummary, dmsSubmission, submission.noOfAttachments)
```
lancewalton: Since we're passing 'submission' here, do we need to pass 'submission.noOfAttachments' too?

josef-vlach: Good catch

PR #492, opened 2019-12-09
GFORMS-291: Provide the ability to have dynamic text in titles, label…

lancewalton:
…s and shortName

### DIFF
--- a/app/uk/gov/hmrc/gform/sharedmodel/SmartString.scala
+++ b/app/uk/gov/hmrc/gform/sharedmodel/SmartString.scala
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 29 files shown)


app/uk/gov/hmrc/gform/sharedmodel/SmartString.scala
```diff
@@ -0,0 +1,118 @@
...
+      localisedString <- readLocalisedStringFromTemplateObject(value)
+      interpolations  <- readInterpolationsFromTemplateObject(value)
+      ss              <- buildSmartStringFromLocalisedStringAndInterpolations(localisedString, interpolations)
+    } yield ss
+
+  val templateReads: Reads[SmartString] = new Reads[SmartString] {
+    override def reads(json: JsValue): JsResult[SmartString] = json match {
+      case JsString(value) => readFromTemplateString(value)
+      case obj: JsObject   => readFromTemplateObject(obj.fields.toMap)
+      case _               => JsError(s"Expected a String or an Object while reading a SmartString. Got a ${json.getClass}")
```
josef-vlach: Maybe output `json` instead of `json.getClass` may help to spot the problem quicker for the user who will see that message

PR #499, opened 2020-01-30
CPDP-1040: draft pdf submissions with attachments

mikail-khan:


app/uk/gov/hmrc/gform/dms/DmsHtmlSubmission.scala
```diff
@@ -18,7 +18,7 @@ package uk.gov.hmrc.gform.dms
 
 import play.api.libs.json.{ Format, Json }
 
-case class DmsHtmlSubmission(html: String, metadata: DmsMetadata)
+case class DmsHtmlSubmission(b64html: String, attachments: List[Array[Byte]], metadata: DmsMetadata)
```
mikail-khan: It seems that the html should be sent in b64 format but that was not clear to me until I dug into the code. So making this explicit in the name here.

app/uk/gov/hmrc/gform/dms/DmsSubmissionAlgebra.scala
```diff
@@ -65,7 +67,28 @@ class DmsSubmissionService[F[_]](
...
+    val formTemplateId = FormTemplateId(metaData.dmsFormId)
+
+    for {
+      pdfBytes   <- pdfGenerator.generatePDFBytes(decode(html))
+      envelopeId <- fileUpload.createEnvelope(formTemplateId, LocalDateTime.now(clock).plusDays(formExpiryDays))
+      pdfDoc = documentLoader(pdfBytes)
+      pdfSummary = PdfSummary(pdfDoc.getNumberOfPages.toLong, pdfBytes)
+      _ = pdfDoc.close()
+      submission = DmsSubmissionService
+        .createSubmission(metaData, envelopeId, LocalDateTime.now(clock), Attachments.empty)
```
mikail-khan: I went through the code and I still can't see how the attachments are processed. All I can see is that it calculates the number of attachments. What about the content? How is that processed and sent down to DMS with the PDF?

app/uk/gov/hmrc/gform/dms/DmsSubmissionWithAttachmentsRequestInterpreter.scala
```diff
@@ -0,0 +1,52 @@
...
+import play.api.libs.json.{ JsError, JsSuccess, Json }
+import play.api.mvc.{ MultipartFormData, Request }
+
+object DmsSubmissionWithAttachmentsRequestInterpreter {
+  def apply(
+    request: Request[MultipartFormData[TemporaryFile]]): Either[String, (String, List[Array[Byte]], DmsMetadata)] = {
+
+    val maybeHtml: Option[Seq[String]] = request.body.dataParts.get("html")
+    val maybeMetadata: Option[DmsMetadata] = Json
+      .toJson(request.body.dataParts.filterKeys(_ == "html").mapValues(_.mkString("")))
```
mikail-khan: this is a bit messy - I am sending the HTML content and the DMS meta data in the dataparts of the MultipartFormData request.
So I filter out the html to just get the DMS meta keys
if you know a better way to do this, let me know

app/uk/gov/hmrc/gform/dms/DmsSubmissionAlgebra.scala
josef-vlach: You need to send attachments into `file-upload-frontend`, that logic is not here...

mikail-khan: :man_facepalming: 

ok mate, let me check it out and see what I need to do - missing jigsaw in the puzzle.

PR #503, opened 2020-02-05
CPDP-1040: remove Java 11 references

mikail-khan:


josef-vlach:
I've applied changes from this PR directly to master, so we have only one commit for CPDP-1040 work.

PR #509, opened 2020-03-16
GFORMS-75 - Updating Welsh in default summary footer to match English

Jonathan-Courage:


Jonathan-Courage:
https://jira.tools.tax.service.gov.uk/browse/GFORMS-75

PR #511, opened 2020-03-31
GFORMS-749 - Header and Footer in Print PDF

SandipKhedekar:


test/uk/gov/hmrc/gform/formtemplate/FormTemplatesControllerRequestHandlerTest.scala
```diff
@@ -344,6 +392,145 @@ class FormTemplatesControllerRequestHandlerTest extends WordSpec with MustMatche
...
+       |            "instructions": "Test Instructions",
+       |            "pdfHeader": "#Test PDF Header",
+       |            "pdfFooter": "##Test PDF Footer"
+       |        },
+       |        "pdf": {
+       |        	"header": "##Test PDF Header",
+       |        	"footer": "##Test PDF Footer",
+       |        	"fieldIds": ["elementA", "elementB"]
+       |        }
+       |    },
```
josef-vlach: Can we change this so that input json looks like this:
```
{
  "page": {
    "title": "SPBP",
    "instructions": "1. Print a copy of the notification"
  },
  "pdfSummary": {
    "header": "[ANY MARKDOWN TEXT]",
    "footer": "[ANY MARKDOWN TEXT]"
  },
  "pdfNotification": {
    "header": "anything",
    "fieldIds": "${field1},${field2}",
    "footer": "anything"
  }
}
```
Why mix what goes into page and with what goes into pdf?

PR #537, opened 2020-09-17
GFORMS-906 Added API to store mtd vat numbers

girishkamat:


josef-vlach (changes requested):
I'm sorry to say that, but I feel this is unnecessary concrete and restricted. 

Uploading payload into mongo collection should be completely generic feature.

Also `allowedListName` and associated type seems to be what we call `collectionName: CollectionName`, see https://github.com/hmrc/gform/blob/master/app/uk/gov/hmrc/gform/dblookup/DbLookupService.scala.

girishkamat:
Thanks @josef-vlach. I had a feeling this would come back. Will take a look at the DbLookup implementation

josef-vlach:
Final goal of this work should be that we will be able to populate new mono collection without need for code change whatsoever.

josef-vlach (approved):

PR #547, opened 2020-10-06
GFORMS-912 - Rewrite BooleanExpr from Equals to Contains for choice/r…

josef-vlach:
…evealing choice

### DIFF
--- a/app/uk/gov/hmrc/gform/ApplicationLoader.scala
+++ b/app/uk/gov/hmrc/gform/ApplicationLoader.scala
@@ -144,8 +144,8 @@ class ApplicationModule(context: Context)
       controllerComponents
     )
 
-  val graphiteModule =
-    new GraphiteModule(environment, configuration, configModule.runMode, applicationLifecycle, metricsModule)
+  // Execute side effects
+  new GraphiteModule(environment, configuration, configModule.runMode, applicationLifecycle, metricsModule)
 
   val dbLookupModule = new DbLookupModule(controllerComponents, mongoModule)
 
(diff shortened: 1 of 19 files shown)


app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -144,8 +144,7 @@ class ApplicationModule(context: Context)
       controllerComponents
     )
 
-  val graphiteModule =
-    new GraphiteModule(environment, configuration, configModule.runMode, applicationLifecycle, metricsModule)
+  new GraphiteModule(environment, configuration, configModule.runMode, applicationLifecycle, metricsModule)
```
girishkamat, line 148: as this is not assigned to any val, do we need this?

app/uk/gov/hmrc/gform/formtemplate/FormTemplateService.scala
```diff
@@ -58,8 +58,8 @@ class FormTemplateService(formTemplateRepo: Repo[FormTemplate], formTemplateRawR
 
   def verifyAndSave(formTemplate: FormTemplate): FOpt[Unit] =
     for {
-      _   <- verify(formTemplate)
-      _   <- formTemplateRepo.upsert(mkSpecimen(formTemplate))
-      res <- formTemplateRepo.upsert(formTemplate)
+      formTemplateToSave <- verify(formTemplate)
+      _                  <- formTemplateRepo.upsert(mkSpecimen(formTemplateToSave))
+      res                <- formTemplateRepo.upsert(formTemplateToSave)
```
girishkamat, line 63: what is the benefit of this change?

app/uk/gov/hmrc/gform/ApplicationLoader.scala
josef-vlach: Good question. I think we need it due to side effects.

girishkamat: Can we add a comment please?

app/uk/gov/hmrc/gform/formtemplate/FormTemplateService.scala
josef-vlach: @girishkamat Is this more clear? It is still not perfect separation, since `rewrite` still does a validation + rewriting. It is so because validation is not doing any type refinement so rewriting would need to do exactly same verification as validation would do.

girishkamat (approved):

PR #549, opened 2020-10-22
GFORMS-679 - Update lookup component to allow filtering of rows displ…

SandipKhedekar:
…ayed by a column value

### DIFF
--- a/app/uk/gov/hmrc/gform/core/parsers/SelectionCriteriaParser.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/SelectionCriteriaParser.scala
@@ -16,6 +16,7 @@
 
 package uk.gov.hmrc.gform.core.parsers
 
+import parseback._
 import uk.gov.hmrc.gform.core.Opt
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ CsvColumnName, FormComponentId, FormCtx, SelectionCriteriaValue }
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.SelectionCriteriaValue._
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/gform/core/parsers/SelectionCriteriaParser.scala
```diff
@@ -0,0 +1,40 @@
...
+object SelectionCriteriaParser {
+
+  def validate(expression: String): Opt[SelectionCriteriaValue] = validateWithParser(expression, selectionCriteria)
+
+  lazy val selectionCriteria = (
+    ExprParsers.expr ^^ { (_, expr) =>
+      SelectionCriteriaExpr(expr)
+    }
+      |
+        FormatParser.fieldId ~ "." ~ FormatParser.alphabeticOnly ^^ { (_, id, _, column) =>
```
josef-vlach, line 33: We should keep consistency and instead of `FormatParser.fieldId` we should use 
```
FormComponentId.unanchoredIdValidation
``` 
as in all other places.

You just need some implicit magic to make that compile, so don't forget to include:
```
import parseback._
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormatExpr.scala
```diff
@@ -19,9 +19,11 @@ package uk.gov.hmrc.gform.sharedmodel.formtemplate
 import cats.Eq
 import julienrf.json.derived
 import play.api.libs.json._
-import uk.gov.hmrc.gform.core.parsers.{ BasicParsers, ValueParser }
+import uk.gov.hmrc.gform.core.parsers.{ BasicParsers, SelectionCriteriaParser, ValueParser }
 import uk.gov.hmrc.gform.sharedmodel.{ EmailVerifierService, LocalisedString }
 
+import scala.collection.immutable.List
```
josef-vlach, line 25: This import is not needed.

josef-vlach (changes requested):

josef-vlach (approved):

PR #555, opened 2020-11-05
GFORMS-961 Support for optionally hiding page titles in add-to-list via presentationHint

girishkamat:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
```diff
@@ -46,11 +48,34 @@ object Section {
...
+        (JsPath \ "includeIf").readNullable[IncludeIf] and
+        (JsPath \ "repeatsMax").readNullable[Expr] and
+        (JsPath \ "pages").read[NonEmptyList[Page]] and
+        (JsPath \ "addAnotherQuestion").read[FormComponent] and
+        (JsPath \ "instruction").readNullable[Instruction] and
+        (JsPath \ "presentationHint").read[Option[List[PresentationHint]]](
+          (json: JsValue) =>
+            parseOpt(JsDefined(json), PresentationHintParser.validate)
+              .fold(us => JsError(us.toString), fv => JsSuccess(fv)))
+    )(AddToList.apply _)
```
josef-vlach, line 72: I don't like this explicit `Reads`. Cannot we just define `Reads[PresentationHint]` which will do the parsing?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
girishkamat: Thanks @josef-vlach. I have replaced this with a `Reads[List[PresentationHint]]`. Have made it a list to mimic what is done on FormComponent level

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
```diff
@@ -46,11 +47,21 @@ object Section {
...
 
+  object AddToList {
+
+    implicit val presentationHintReads: Reads[List[PresentationHint]] = (json: JsValue) =>
+      parseOpt(JsDefined(json), PresentationHintParser.validate)
+        .fold(us => JsError(us.toString), fv => JsSuccess(fv.getOrElse(List.empty)))
+
+    implicit val format: Format[AddToList] = Json.format[AddToList]
+  }
+
```
josef-vlach, line 64: I was hoping in more subtle solution, like hiding this `Reads[List[PresentationHint]]` in `def readAddToList(json: JsValue)` and not implementing it in terms of `parseOpt`.

Something like:

```
private def readAddToList(json: JsValue) = {
  implicit def presentationHintsReads: OFormat[List[PresentationHint]] =
    OFormatWithTemplateReadFallback {
      case JsString(str) => PresentationHintParser.validate(str).fold(e => JsError(e.error), JsSuccess(_))
      case unknown       => JsError("Expected String, got " + unknown)
    }
  Json.format[Section.AddToList].reads(json)
}
```

Then we can revert `toOpt` changes etc.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
josef-vlach: This is still a dirty solution, though. The best thing would be to introduce newtype for `List[PresentationHint]` and do the logic there.

girishkamat: Implemented as per the discussion

josef-vlach (approved):

PR #561, opened 2020-11-23
GFORMS-986 format is mandatory for form components of type text

girishkamat:


test/uk/gov/hmrc/gform/sharedmodel/FieldValueFormatValueSpec.scala
```diff
@@ -27,13 +27,14 @@ class FieldValueFormatValueSpec extends Spec {
...
-           "value": "${amountA + amountB}"
+           "value": "${amountA + amountB}",
+           "format": "text"
           }
       """)
     fieldValue should beJsSuccess(
       FormComponent(
         FormComponentId("sum"),
-        Text(ShortText.default, Add(FormCtx(FormComponentId("amountA")), FormCtx(FormComponentId("amountB")))),
+        Text(BasicText, Add(FormCtx(FormComponentId("amountA")), FormCtx(FormComponentId("amountB")))),
```
josef-vlach, line 37: `BasicText` is weird beast, do we really want to keep it?

 It is never used with `Text` as it is only constructed for multilines. Maybe we should mandate `format` on multiline as well and get rid of `BasicText`?

Note: It looks like `BasicText` is the same as `format: "text(0, 100000)", we should get rid of it.

girishkamat: Alex is fine with making it mandatory for TextArea. Will get rid of BasicText then. So for testing, format=text(0, 100000) should be ok then?

girishkamat: Removed BasicText and also updated docs at https://confluence.tools.tax.service.gov.uk/display/DF/Gforms+Text+Formats

app/uk/gov/hmrc/gform/core/parsers/FormatParser.scala
```diff
@@ -127,8 +127,6 @@ object FormatParser {
         TextFormat(ShortText.default)
       } | "shortText(" ~ positiveInteger ~ "," ~ positiveInteger ~ ")" ^^ { (_, _, min, _, max, _) =>
         TextFormat(ShortText(min, max))
-      } | "text" ^^ { (_, _) =>
```
josef-vlach, line 130: This is removing little to much.
I think this parser should stay, it should only produce `TextFormat(TextWithRestrictions(0, 100000))`.
Or did you agree with Alex that `"format":"text"` shouldn't be used?
Also removing of `BasicText` needs to propagate to frontend.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/core/parsers/FormatParser.scala
girishkamat: I haven't agreed anything with Alex. I thought we were supposed to get rid of text. I will map it to TextFormat(TextWithRestrictions(0, 100000)), as ppl are used to that term

girishkamat: "text" has been restored and mapped to `TextFormat(TextWithRestrictions(0, 100000))`

girishkamat: Will create fontend PR for removal of BasicText

josef-vlach (approved):

PR #569, opened 2020-12-09
508 patch

girishkamat:


josef-vlach:
Since there is #570 and #567 I don't think we need this.

girishkamat:
You are right

PR #576, opened 2020-12-24
GFORMS-890 Upgrading Play to 2.7

girishkamat:


SandipKhedekar (approved):

app/uk/gov/hmrc/gform/commons/HttpFunctions.scala
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.gform.commons
+
+import uk.gov.hmrc.http.HttpReads.{ is2xx, is4xx, is5xx }
+import uk.gov.hmrc.http._
+
+trait HttpFunctions {
+  def jsonHttpReads[A](rds: HttpReads[A]): HttpReads[A] =
```
josef-vlach, line 23: Why is this functionality implemented here? Shouldn't this be part of some platform-wide library?

girishkamat: I checked the docs, the new implementation doesn't exactly replicate the old deprecated reader and expects us to implement custom one if we need the exact behaviour as before. The old implementation looks tailored to a specific use case. 

Personally, I would expect the framework to do minimal when it comes to response processing and let the application decide what to do

josef-vlach: But why we want to stick with an old implementation? This `HttpReads` this builds on seems for some legacy purposes. Why not to switch to non-legacy way of doing http calls?

josef-vlach: By that I mean whats wrong with using
```
import uk.gov.hmrc.http.HttpReads.Implicits._
```
instead of this implementation?

girishkamat: We use BadRequestException and NotFoundException in other layers. Using the implicits means updating the handling of exceptions in the top layers. These exceptions are also being thrown in other areas of code.

josef-vlach: > Using the implicits means updating the handling of exceptions in the top layers

This is one place to update (perhaps couple of lines of code) vs using this wrapper anytime we want to make a call.

> These exceptions are also being thrown in other areas of code.

I think this is ok, no need to do anything about this. This is client duty to handle appropriately isn't it?

girishkamat: Ok, will take a look

josef-vlach (approved):

PR #582, opened 2021-01-20
GFORMS-871 - Add the ability to specify the number of rows in a multi…

SandipKhedekar:
…-line text field

### DIFF
--- a/app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
@@ -102,7 +102,7 @@ object BasicParsers {
 
   lazy val positiveInteger: Parser[Int] = intParser("""\d+""")
 
-  lazy val nonZeroPositiveInteger: Parser[Int] = intParser("""^\d*[1-9]\d*$""")
+  lazy val nonZeroPositiveInteger: Parser[Int] = intParser("""[1-9]{1}\d*""")
 
   lazy val anyInteger: Parser[Int] = intParser("""(\+|-)?\d+""")
 
--- a/app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
@@ -51,6 +51,9 @@ object BasicParsers {
       expr      <- reconstruct(catenable)
     } yield expr).run(expression.trim)
 
+  def validateNonZeroPositiveNumber(expression: Int): Opt[Int] =
+    validateWithParser(expression.toString, nonZeroPositiveInteger)
+
   implicit val W = Whitespace(() | """\s+""".r)
 
   def nextOrPreviousValue[A](string: String, fn: (Int, Int) => A): Parser[A] =
@@ -98,6 +101,8 @@ object BasicParsers {
   lazy val exactDayParser: Parser[Int] = intParser("""0[1-9]|[12][0-9]|3[01]""")
 
   lazy val positiveInteger: Parser[Int] = intParser("""\d+""")
+
+  lazy val nonZeroPositiveInteger: Parser[Int] = intParser("""^\d*[1-9]\d*$""")
 
   lazy val anyInteger: Parser[Int] = intParser("""(\+|-)?\d+""")
 
--- a/app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
@@ -102,7 +102,7 @@ object BasicParsers {
 
   lazy val positiveInteger: Parser[Int] = intParser("""\d+""")
 
-  lazy val nonZeroPositiveInteger: Parser[Int] = intParser("""[1-9]{1}\d*""")
+  lazy val nonZeroPositiveInteger: Parser[Int] = intParser("""^\d*[1-9]\d*$""")
 
   lazy val anyInteger: Parser[Int] = intParser("""(\+|-)?\d+""")
 
--- a/app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
+++ b/app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
@@ -229,7 +229,7 @@ class FormComponentMaker(json: JsValue) {
                  toUpperCase,
                  prefix,
                  suffix,
-                 rows,
+                 rows.getOrElse(TextArea.defaultRows),
                  json)
     } yield result
   }
--- a/app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
+++ b/app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
@@ -89,6 +89,7 @@ class FormComponentMaker(json: JsValue) {
     toOpt((json \ "validators").validateOpt[List[FormComponentValidator]], "/validators").map(_.toList.flatten)
   lazy val mandatory: Option[String] = (json \ "mandatory").asOpt[String]
   lazy val multiline: Option[String] = (json \ "multiline").asOpt[String]
+  lazy val rows: Option[Int] = (json \ "rows").asOpt[Int]
   lazy val displayWidth: Option[String] = (json \ "displayWidth").asOpt[String]
   lazy val toUpperCase: UpperCaseBoolean = (json \ "toUpperCase").asOpt[UpperCaseBoolean].getOrElse(IsNotUpperCase)
   lazy val prefix: Option[SmartString] = (json \ "prefix").asOpt[SmartString]
@@ -227,6 +228,7 @@ class FormComponentMaker(json: JsValue) {
                  toUpperCase,
                  prefix,
                  suffix,
+                 rows,
                  json)
     } yield result
   }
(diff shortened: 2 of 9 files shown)


app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
```diff
@@ -89,6 +89,7 @@ class FormComponentMaker(json: JsValue) {
     toOpt((json \ "validators").validateOpt[List[FormComponentValidator]], "/validators").map(_.toList.flatten)
   lazy val mandatory: Option[String] = (json \ "mandatory").asOpt[String]
   lazy val multiline: Option[String] = (json \ "multiline").asOpt[String]
+  lazy val rows: Option[Int] = (json \ "rows").asOpt[Int]
```
josef-vlach, line 92: I believe there exists some reasonable default here, can we make `rows` an `Int` instead of an `Option[Int]`?

Or let's wait on frontend implementation instead, to see how well this plays with it.

test/uk/gov/hmrc/gform/sharedmodel/JsonParseTestFormat.scala
```diff
@@ -181,4 +181,52 @@ class JsonParseTestFormat extends Spec with TableDrivenPropertyChecks {
     }
   }
 
+  "A text area component with a valid no of rows" should "parse correctly" in {
+
+    val multilineCombinations = Table(
```
girishkamat, line 186: dont think we need Table driven tests here. You are not testing any variation. A simple example with 1 or 11 rows is good enough. Avoid magic numbers like 7, 10 , 40 etc

test/uk/gov/hmrc/gform/sharedmodel/JsonParseTestFormat.scala
```diff
@@ -181,4 +181,52 @@ class JsonParseTestFormat extends Spec with TableDrivenPropertyChecks {
...
+      jsResult shouldBe a[JsSuccess[_]]
+      jsResult.map(fv => fv.`type` shouldBe expected)
+
+    }
+  }
+
+  "A text area component with an invalid no of rows" should "default to text area" in {
+
+    for {
+      snippet <- List(""", "format": "shortText", "multiline" : "yes", "rows" : "INVALID NO OF ROWS"}""")
```
girishkamat, line 207: Why is this a list?

girishkamat: Btw,  arent we expected to get an error instead of silently defaulting to None?

test/uk/gov/hmrc/gform/sharedmodel/formtemplate/generators/ComponentTypeGen.scala
```diff
@@ -24,6 +24,7 @@ import uk.gov.hmrc.gform.sharedmodel.formtemplate._
 trait ComponentTypeGen {
 
   def displayWidthGen: Gen[DisplayWidth] = Gen.oneOf(DisplayWidth.values.toSeq)
+  def rowsGen: Gen[Option[Int]] = Gen.option(Gen.chooseNum(5, 50))
```
girishkamat, line 27: For test purpose, we dont need to restrict to 5-50. Any positive number will do.

test/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormComponentMakerServiceSpec.scala
```diff
@@ -137,7 +146,16 @@ class FormComponentMakerServiceSpec extends Spec with TableDrivenPropertyChecks
   it should "return error when format is empty for multiline text type" in {
     val isMultiline = Some("yes") // denotes multiline text type
     val result =
```
girishkamat, line 148: missing test when rows are specified in createObject

app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
SandipKhedekar: In a library **play-frontend**, component **Textarea** **rows** has been defaulted to 5. As per discussion with Alex, if we don't provide rows in a template then we shouldn't provide any value of rows to library.

josef-vlach: Fair enough

josef-vlach: But actually I'm still interested, how is this `Option` beneficial? How are you avoid calling `.getOrElse` on it to provide `Int` value to the **play-frontend** ?

SandipKhedekar: Will use **fold** on **rows: Option[Int]** with default value without row parameter.

test/uk/gov/hmrc/gform/sharedmodel/JsonParseTestFormat.scala
SandipKhedekar: Have changed to a single value test. But there is no harm to test with multiple values.

test/uk/gov/hmrc/gform/sharedmodel/JsonParseTestFormat.scala
SandipKhedekar: Removed List. There is no error in case of invalid value for rows same as **multiline**. It will default to None.

test/uk/gov/hmrc/gform/sharedmodel/formtemplate/generators/ComponentTypeGen.scala
SandipKhedekar: Changed to **posNum**.

girishkamat: the value generated is irrelevant, as long its a positive number

test/uk/gov/hmrc/gform/sharedmodel/JsonParseTestFormat.scala
girishkamat: No, but while reading it can get confusing when such magic numbers are seen. Does it only work for these 3 numbers. As far as possible, stick to simple values like 1, A etc

app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
girishkamat: I prefer making it a Int instead of Optional, especially because we have a way to add  the default value to JSON if it doesnt exist

app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
```diff
@@ -99,6 +102,8 @@ object BasicParsers {
 
   lazy val positiveInteger: Parser[Int] = intParser("""\d+""")
 
+  lazy val nonZeroPositiveInteger: Parser[Int] = intParser("""^\d*[1-9]\d*$""")
```
girishkamat, line 105: could this be simplified - `"""[1-9]{1}\d*"""`

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
```diff
@@ -106,6 +107,10 @@ object DisplayWidth extends Enumeration {
   implicit val displayWidthWrites: Writes[DisplayWidth] = Writes.enumNameWrites
 }
 
+object TextAreaRows {
```
girishkamat, line 110: I think we can keep this object name as TextArea, no need to define another object just for it

`
object TextArea {
   val default = 5
}
`

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
```diff
@@ -57,7 +57,8 @@ object UpperCaseBoolean {
 case class TextArea(
   constraint: TextConstraint,
   value: Expr,
-  displayWidth: DisplayWidth = DisplayWidth.DEFAULT
+  displayWidth: DisplayWidth = DisplayWidth.DEFAULT,
+  rows: Int = TextAreaRows.default
```
girishkamat, line 61: Is it required to specify the default here?

app/uk/gov/hmrc/gform/formtemplate/FormComponentMakerService.scala
```diff
@@ -63,11 +64,14 @@ object FormComponentMakerService {
     maybeValueExpr: Option[ValueExpr],
     displayWidth: Option[String],
     multiLine: Option[String],
+    rows: Option[Int],
```
josef-vlach, line 67: This may be already an `Int`. 

It will avoid repetition of `rows.getOrElse(TextAreaRows.default)` down the line.

app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
SandipKhedekar: Done.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
SandipKhedekar: Done.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
SandipKhedekar: Yes. I have used default argument value at many places in tests.

app/uk/gov/hmrc/gform/formtemplate/FormComponentMakerService.scala
SandipKhedekar: Perfect. Done.

app/uk/gov/hmrc/gform/core/parsers/BasicParsers.scala
girishkamat: Actually realized, this would not allow 01 and also match on non-numbers. I think your solution is the best one. Can you please revert?

girishkamat (approved):

PR #592, opened 2021-02-05
GFORMS-1069 - A property to allow multiline text fields without a cha…

SandipKhedekar:
…racter count


josef-vlach:
I'll look at this once there is PR for frontend part.

josef-vlach (approved):

PR #593, opened 2021-02-08
GFORMS-1076 - Add a new expression to allow for the count of Add to

josef-vlach:
List items to be displayed

Allow ${addAnother.count} expression in value field

### DIFF
--- a/app/uk/gov/hmrc/gform/formtemplate/SectionHelper.scala
+++ b/app/uk/gov/hmrc/gform/formtemplate/SectionHelper.scala
@@ -17,7 +17,7 @@
 package uk.gov.hmrc.gform.formtemplate
 
 import cats.data.NonEmptyList
-import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ Page, Section }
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/gform/formtemplate/SectionHelper.scala
```diff
@@ -29,4 +29,13 @@ object SectionHelper {
 
   def pages(sections: List[Section]): List[Page] =
     sections.flatMap(pages(_).toList)
+
+  def addToListRepeater(section: Section): Option[FormComponent] =
```
girishkamat, line 33: Just curious why we don't have Section.fold, like we do in other cases?

josef-vlach: Because backend is not so polished

girishkamat: ok

girishkamat (approved):

PR #595, opened 2021-02-12
GFORMS-790 - Enhance choice component to allow hints against the choices

josef-vlach:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ComponentType.scala
```diff
@@ -118,6 +118,7 @@ case class Choice(
   options: NonEmptyList[SmartString],
   orientation: Orientation,
   selections: List[Int],
+  hints: Option[NonEmptyList[SmartString]],
```
girishkamat, line 121: Why done we group options and hints together as a single object and make hint optional if needed? This will avoid the need for validations and looks more practical than keeping them separate. We can always pre-process the json for backwards compatibility no?

josef-vlach: Yes, option, hint and optionHelpText should be one object ideally, but that is out of scope for GFORMS-790.

girishkamat: Ok

girishkamat (approved):

PR #606, opened 2021-03-09
Update "sbt-scalafmt" from "com.lucidchart" to "org.scalameta"

josef-vlach:


.scalafmt.conf
```diff
@@ -1,47 +1,31 @@
...
+continuationIndent.callSite = 2
+continuationIndent.defnSite = 2
+
+newlines.penalizeSingleSelectMultiArgList = false
+newlines.sometimesBeforeColonInMethodReturnType = true
+
+
+rewrite.rules = [RedundantBraces, RedundantParens, AsciiSortImports]
+rewrite.redundantBraces.maxLines = 120
+rewrite.redundantBraces.includeUnitMethods = true
```
SandipKhedekar, line 25: Is it formatted by **scalafmt** to flat instead of tree structure?

josef-vlach: No, I rewrote that. 

It looks like it is preferred syntax: https://scalameta.org/scalafmt/docs/configuration.html#other

SandipKhedekar (approved):
LGTM

PR #624, opened 2021-05-07
GFORMS-1208 - UK Address component SmartString

josef-vlach:


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -184,13 +184,24 @@ object ValueParser {
     | FormComponentId.unanchoredIdValidation ~ ".count" ^^ { (loc, value, _) =>
       Count(FormComponentId(value))
     }
+    | FormComponentId.unanchoredIdValidation ~ "." ~ addressDetail ^^ { (loc, value, _, addressDetail) =>
+      AddressLens(FormComponentId(value), addressDetail)
```
girishkamat, line 188: Why dont we extend the FormCtx, to include a optional detail attribute.

`FormCtx(formComponentId: FormComponentId, detail: Option[String] = None) extends Expr
`
While saving, we can validate what detail is allowed for a given form component i.e for Address - street1, street2 etc`

On the gform-frontend, we just need to do

`formCtx.detail.fold(expr.formComponentId.modelComponentId)(d => formCtx.formComponentId.modelComponentId.toAtomicFormComponentId(Atom(d))
    )`

josef-vlach: @girishkamat you can use triple backtick: `` ``` ``
To format multiline code:
```
formCtx.detail.fold(expr.formComponentId.modelComponentId)(d => 
formCtx.formComponentId.modelComponentId.toAtomicFormComponentId(Atom(d)) )
```

josef-vlach: Having `detail` as an  `Option[String] = None` would be quite unsafe, since `String` is not constrained in any way. 
On the other hand dedicated class like:
```
AddressLens(formComponentId: FormComponentId, detail: AddressDetail)
```
is much more safe to use.

Also having defined `FormCtx` like:

```
FormCtx(formComponentId: FormComponentId, detail: Option[String] = None) extends Expr
```

Would make it confusing as to what `detail` is for. In most places it would be not used. Basically connection between `Address` component and its `detail` would be lost (it won't be obvious when reading such definition).

Last but not least having `FormCtx` with `detail` would make `FormCtx` inextensible. What if we want to store any other information for `Address`? We can easily extend `AddressLens` however we want without regards to the rest of `Expr` family, but that wouldn't be the case with `FormCtx` with `detail`.

PR #626, opened 2021-05-28
GFORMS-1073 - Group component is not allowed in Declaration section

SandipKhedekar:


app/uk/gov/hmrc/gform/formtemplate/FormTemplatesControllerRequestHandler.scala
```diff
@@ -195,7 +199,13 @@ object FormTemplatesControllerRequestHandler {
...
       }
 
-    sectionValidations andKeep dmsFormIdValidations andKeep jsonValue.transform(
+    val declarationSectionValidations: JsResult[Unit] = (jsonValue \ "declarationSection" \ "fields").toOption match {
+      case Some(fields) if fields.as[JsArray].value.flatMap(jv => (jv \ "type").asOpt[String]).contains("group") =>
+        invalidGroupComponentInDeclarationSection
+      case _ => JsSuccess(())
+    }
+
+    declarationSectionValidations andKeep sectionValidations andKeep dmsFormIdValidations andKeep jsonValue.transform(
```
josef-vlach, line 208: We should not validate this on json level. Json level validation are supposed to be really low level.

We should check this on `FormTemplate` level. Best place is to add new check into [Verifier.scala](https://github.com/hmrc/gform/blob/master/app/uk/gov/hmrc/gform/formtemplate/Verifier.scala)

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/formtemplate/DestinationsValidator.scala
```diff
@@ -36,15 +39,25 @@ object DestinationsValidator {
...
+      val groupComponentIds = extractGroupComponentIds(destinationList.declarationSection.fields)
+      groupComponentIds.isEmpty.validationResult(groupComponentInDeclaration(groupComponentIds))
+  }
+
+  def extractGroupComponentIds(fcs: List[FormComponent]): List[FormComponentId] = fcs
+    .map(fc => (fc.id, fc.`type`))
+    .collect { case (fcId, Group(gFields, _, _, _, _)) =>
+      fcId :: extractGroupComponentIds(gFields)
+    }
+    .flatten
```
josef-vlach, line 54: We can make this simpler, by returning only `Option[FormComponentId]`. 

Also recursion can be eliminated, since outer `Group` will be always detected.

Also use existing `IsGroup` extractor rather than custom `map(...).collect(...)` combo. My guess would be to implemented it somehow like `fcs.collectFirst { case g @ IsGroup(_)) => g.id }`

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/formtemplate/DestinationsValidator.scala
```diff
@@ -18,14 +18,17 @@ package uk.gov.hmrc.gform.formtemplate
...
+import uk.gov.hmrc.gform.core.{ Valid, ValidationResult }
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ FormComponent, FormComponentId, IsGroup }
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.destinations.{ Destination, DestinationId, Destinations }
 
 object DestinationsValidator {
   def someDestinationIdsAreUsedMoreThanOnce(duplicates: Set[DestinationId]) =
     s"Some DestinationIds are defined more than once: ${duplicates.toList.sortBy(_.id).map(_.id)}"
 
+  def groupComponentInDeclaration(groupComponentId: Option[FormComponentId]) =
+    groupComponentId.fold("")(id => s"Group component $id is not allowed in Declaration")
```
josef-vlach, line 30: From reader point of view this looks weird, why is `Option[FormComponentId]` passed here and what empty String as a return value means?

D I bet there is better way how to express this hint. use `fold` instead of `exists` later on?

josef-vlach (changes requested):

josef-vlach (approved):

PR #627, opened 2021-06-01
GFORMS-281 Verify email template id and email parameters

girishkamat:


josef-vlach:
Interesting, the intention was to have this check as part of deployment process in eeitt-admin-frontend.

Do we really want to force people to have valid email template when they start to work on new form? I'd bet email-template creating come much later in production process. So why bother people with that at all times?

girishkamat:
As discussed, moving to eeitt-admin-frontend

PR #632, opened 2021-06-11
GFORMS-1309 Implemented extensions to period function

girishkamat:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
```diff
@@ -40,6 +40,17 @@ final case class FormCtx(formComponentId: FormComponentId) extends Expr
 final case class AddressLens(formComponentId: FormComponentId, detail: AddressDetail) extends Expr
 final case class PeriodFun(dateCtx1: Expr, dateCtx2: Expr) extends Expr
 
+sealed trait FunctionProp
```
josef-vlach, line 43: `FunctionProp `sounds more generic than it actually is. Why not to call it `PeriodFunctionProp`? Or perhaps just `PeriodFn`?

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
```diff
@@ -40,6 +40,17 @@ final case class FormCtx(formComponentId: FormComponentId) extends Expr
...
+case object SumProp extends FunctionProp
+case object TotalMonthsProp extends FunctionProp
+case object YearsProp extends FunctionProp
+case object MonthsProp extends FunctionProp
+case object DaysProp extends FunctionProp
```
josef-vlach, line 48: These definition would better be placed inside `object FunctionProp` to give them proper namespace. Also they would not need to ends with `Prop`.

If this and above suggestions were implemented, then referring to these objects would look like this:

- `PeriodFn.Sum`
- `PeriodFn.TotalMonths`
- `PeriodFn.Years`

etc.

girishkamat: We already have case class PeriodFun. PeriodFn sounded very similar, thats why I was thinking of Prop, meaning property of the function. 

May be I can change the original case class to Period?

```
final case class Period(dateCtx1: Expr, dateCtx2: Expr) extends Expr
sealed trait PeriodFn
object PeriodFn {
  case object Sum extends PeriodFn
  case object TotalMonths extends PeriodFn
  case object Years extends PeriodFn
  case object Monthsextends PeriodFn
  case object Days extends PeriodFn
}

final case class PeriodExt(periodFun: Expr, func: PeriodFn) extends Expr
```

Thoughts?

josef-vlach: What about renaming `PeriodFun` to just `Period`? Clashes between this and java `Period` would be minimal I believe.

josef-vlach: Yes, that makes sense. 👍

josef-vlach (approved):

PR #634, opened 2021-06-17
GFORMS-1326 - Gform log error on create Envelope failure

SandipKhedekar:


app/uk/gov/hmrc/gform/fileupload/FileUploadConnector.scala
```diff
@@ -45,9 +45,23 @@ class FileUploadConnector(config: FUConfig, wSHttp: WSHttp, timeProvider: TimePr
...
+        val status = response.status
+        if (status >= 200 && status < 300) {
+          Future.successful(helper.extractEnvelopId(response))
+        } else {
+          Future.failed(
+            new Exception(s"POST to $url failed with status $status. Response body: '${response.body}'")
+          )
+        }
+      } recoverWith { case ex =>
+      Future.failed(new Exception(s"POST to $url failed. $ex"))
```
girishkamat, line 63: Just out of curiosity, what was the type of ex before? Why wasnt it working?

SandipKhedekar: It was **uk.gov.hmrc.http.Upstream4xxResponse** with 404 code. On this line, https://github.com/hmrc/http-verbs/blob/master/http-verbs-common/src/main/scala/uk/gov/hmrc/http/logging/ConnectionTracing.scala#L41 in library it's just logging as INFO. And in our code, POST call to File upload service was failing and future recover/recoverWith was missing. Hence we were not logging as ERROR for exception and missing alerts.

girishkamat: Ok, not sure if 404 is the right code that file-upload should return. This is obviously a bad request i.e max files is incorrect. Anyways, if the API is returning 404, there is nothing much you could do about it

girishkamat (approved):

PR #638, opened 2021-07-07
GFORMS-1140 Using MongoCaheRepo to save data in local mongo collection

girishkamat:


app/uk/gov/hmrc/gform/save4later/EncyryptedFormat.scala
```diff
@@ -0,0 +1,37 @@
...
+
+package uk.gov.hmrc.gform.save4later
+
+import play.api.libs.json._
+import uk.gov.hmrc.crypto.{ Crypted, CryptoWithKeysFromConfig, PlainText }
+
+import scala.util.{ Failure, Success, Try }
+
+object EncyryptedFormat {
+  def formatEncrypted[A](jsonCrypto: CryptoWithKeysFromConfig)(implicit format: Format[A]): Format[A] = new Format[A] {
```
josef-vlach, line 25: Why is this encryption needed?

girishkamat: Form data may contain sensitive data. Its a good practice to encrypt data at REST. Save4later was doing this as well

josef-vlach: To enable data encryption was the reason why `save4later` was build. That was in the days mongo connection was not using https. But since https was introduced years ago and gform runs in protected zone I don't think we need to encrypt our data on top of all of that.

girishkamat: Ok, as forms contains PII and mongo db is not password protected, someone with access to the protected zone can still access data in the DB. I will drop a message on #team-platsec and confirm if its needed

test/resources/logback-test.xml
```diff
@@ -97,7 +97,7 @@
         <appender-ref ref="STDOUT"/>
     </logger>
 
-    <root level="ERROR">
+    <root level="INFO">
```
josef-vlach, line 100: Can we keep this at ERROR.

app/uk/gov/hmrc/gform/save4later/FormCacheWithFallback.scala
```diff
@@ -0,0 +1,61 @@
...
+        logger.info(s"Form not found in local mongo cache. Checking save4later [formId=${formId.value}]")
+        save4Later.find(formId)
+      } { form =>
+        Future.successful(Some(form))
+      }
+    }
+
+  override def get(formId: FormId)(implicit hc: HeaderCarrier): Future[Form] = find(formId) map {
+    case None =>
+      throw UpstreamErrorResponse(
```
josef-vlach, line 45: Why not return `Future.failed` ?

app/uk/gov/hmrc/gform/save4later/FormCacheWithFallback.scala
```diff
@@ -0,0 +1,61 @@
...
+        s"Not found 'form' for the given id: '${formId.value}'",
+        StatusCodes.NotFound.intValue
+      )
+    case Some(form) =>
+      logger.debug(Json.prettyPrint(Json.toJson(form)) + "GETFORM")
```
josef-vlach, line 50: Does anyone need this debug message?

app/uk/gov/hmrc/gform/save4later/FormMongoCache.scala
```diff
@@ -0,0 +1,62 @@
...
+  private val formDataKey: DataKey[Form] = DataKey("form")
+
+  private val logger: Logger = LoggerFactory.getLogger(getClass)
+
+  override def find(formId: FormId)(implicit hc: HeaderCarrier): Future[Option[Form]] = mongoCacheRepository
+    .get[Form](formId.value)(formDataKey)
+
+  override def get(formId: FormId)(implicit hc: HeaderCarrier): Future[Form] = find(formId) map {
+    case None =>
+      throw UpstreamErrorResponse(
```
josef-vlach, line 44: Same as above

app/uk/gov/hmrc/gform/save4later/FormMongoCache.scala
```diff
@@ -0,0 +1,62 @@
...
+        s"Not found 'form' for the given id: '${formId.value}'",
+        StatusCodes.NotFound.intValue
+      )
+    case Some(form) =>
+      logger.debug(Json.prettyPrint(Json.toJson(form)) + "GETFORM")
```
josef-vlach, line 49: Same as above

app/uk/gov/hmrc/gform/save4later/FormMongoCache.scala
```diff
@@ -0,0 +1,62 @@
...
+      form
+  }
+
+  override def get(formIdData: FormIdData)(implicit hc: HeaderCarrier): Future[Form] = get(formIdData.toFormId)
+
+  override def upsert(formId: FormId, form: Form)(implicit hc: HeaderCarrier): Future[Unit] =
+    mongoCacheRepository
+      .put(formId.value)(formDataKey, form)
+      .map(_ => ())
```
josef-vlach, line 58: Is this safe, what if write failed?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/save4later/FormCacheWithFallback.scala
girishkamat: Is there an issue doing this way? To return `Future.failed`, I need to flatMap and for success wrap in `Future.successful`

app/uk/gov/hmrc/gform/save4later/FormCacheWithFallback.scala
girishkamat: Probably not, will remove

app/uk/gov/hmrc/gform/save4later/FormMongoCache.scala
girishkamat: `put` returns a CacheItem, which contains details of the object insert in a success scenario, which we dont need. All failures will throw exception and cause Future to fail. We dont need anything from CacheItem

test/resources/logback-test.xml
girishkamat: Done

app/uk/gov/hmrc/gform/save4later/FormMongoCache.scala
josef-vlach: Ok

josef-vlach (approved):

PR #682, opened 2021-09-08
GFORMS-1377 - Fixes

SandipKhedekar:


josef-vlach:
Fixes is not much useful. Would there be some other message to use instead "Fixes"?

josef-vlach (approved):

PR #684, opened 2021-09-13
GFORMS-1466 Implemented expression user.enrolments.[SERVICE].[IDENTIFIER].count/index

girishkamat:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
```diff
@@ -103,6 +103,7 @@ object FormCtx {
 final case class ParamCtx(queryParam: QueryParam) extends Expr
 final case class AuthCtx(value: AuthInfo) extends Expr
 final case class UserCtx(value: UserField) extends Expr
+final case class UserFuncCtx(value: UserField, func: UserFieldFunc) extends Expr
```
josef-vlach, line 106: This assume that `func` can be applied to any product from `value` type. Is that correct? For example do we want to ask `affinityGroup.count` or `enrolledIdentifier.count`?

app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -161,7 +163,15 @@ object ValueParser {
 
   private lazy val periodFun = "period(" ~ dateExpr ~ "," ~ dateExpr ~ ")"
 
-  lazy val contextField: Parser[Expr] = ("user" ~ "." ~ userField ^^ { (loc, _, _, userField) =>
+  private lazy val userEnrolmentFunc = "user" ~ "." ~ userFieldEnrolments ~ "." ~ "(count|[1-9][0-9]*)".r ^^ {
```
josef-vlach, line 166: Can we turn `(count|[1-9][0-9]*)` into `Parse[UserFuncCtx]`? Also we already have `intParser` which can be used here.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala
girishkamat: It does say so looking at the type definition. I dont about AffinityGroup or EnrolledIdentifier. Alternatively, to be precise, we could rename the class to

```final case class UserEnrollmentFuncCtx(enrolment: UserField.Enrolment, func: UserFieldFunc) extends Expr```

Comment (girishkamat) [reply app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala:106]: Actually, when using the above, derived is unable to build a OFormat[_] for UserField class. This has happened in the past for other Expr types too

```
/Users/girishkamath/development/hmrc/gform/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala:117:55
could not find Lazy implicit value of type julienrf.json.derived.DerivedReads[A]
Error occurred in an application involving default arguments.
  implicit val format: OFormat[Expr] = derived.oformat()
```

File: app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala (hunk 2 above)
Comment (girishkamat) [reply app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala:166]: Yes, definitely

File: app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala (hunk 1 above)
Comment (josef-vlach) [reply app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala:106]: Problems aside I'm not in favour of this definition (ie. defining new `Expr` for something which is not new) 
```
final case class UserEnrollmentFuncCtx(enrolment: UserField.Enrolment, func: UserFieldFunc)
```
I'd rather express it like this
```
final case class Enrolment(serviceName: ServiceName, identifierName: IdentifierName, func: UserFieldFunc)
      extends UserField
```
We can introduce synthetic function `Print` to express situation when no function is defined.

Comment (girishkamat) [reply app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala:106]: Why not introduce a new UserField and avoid Print alltogether. This synthetic function is being forced upon with the approach

```final case class EnrolmentFunc(serviceName: ServiceName, identifierName: IdentifierName, func: UserFieldFunc)
      extends UserField```

or 

```final case class EnrolmentFunc(enrolment: UserField.Enrolment, func: UserFieldFunc)
      extends UserField```

Comment (josef-vlach) [reply app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala:106]: Introducing new `UserField` is introducing more friction. 

And in such a case whole `final case class Enrolment` would be synthetic.

If you don't like synthetic function representation `case object Print extends UserFieldFunc` we can be less expressive and make `UserFieldFunc` optional in `Enrolment` instead:
```
final case class Enrolment(serviceName: ServiceName, identifierName: IdentifierName, func: Option[UserFieldFunc])
      extends UserField
```

Comment (girishkamat) [reply app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Expr.scala]: Thanks, I have made it optional. Please re-review

File: app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala [hunk 3]
```diff
@@ -353,4 +369,8 @@ object ValueParser {
   )
 
   private def const[A](a: A)(loc: List[Line], matched: String): A = a
+
+  object Int {
```
josef-vlach, line 373: This is unused, let's remove it.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
girishkamat: done

josef-vlach (changes requested):
This code doesn't compile as it stands.

(Github quick edit changes are always backfiring I guess)

girishkamat:
> This code doesn't compile as it stands.
> 
> (Github quick edit changes are always backfiring I guess)

Haha....it was done in the IDE and pushed. However didnt remove the unwanted import

josef-vlach (approved):

PR #686, opened 2021-09-20
GFORMS-1444 - Add file-upload component flag for determining upload p…

josef-vlach:
…rocess upscan/file-upload

### DIFF
--- a/app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
+++ b/app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
@@ -62,6 +62,14 @@ class FormComponentMaker(json: JsValue) {
     case JsUndefined() => Right(EmailVerification.noVerification)
   }
 
+  lazy val optFileUploadProvider: Opt[FileUploadProvider] = (json \ "service") match {
+    case JsDefined(JsString("upscan"))     => Right(FileUploadProvider.Upscan)
+    case JsDefined(JsString("fileUpload")) => Right(FileUploadProvider.FileUploadFrontend)
+    case JsDefined(unknown) =>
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
```diff
@@ -62,6 +62,14 @@ class FormComponentMaker(json: JsValue) {
...
   }
 
+  lazy val optFileUploadProvider: Opt[FileUploadProvider] = (json \ "service") match {
+    case JsDefined(JsString("upscan"))     => Right(FileUploadProvider.Upscan)
+    case JsDefined(JsString("fileUpload")) => Right(FileUploadProvider.FileUploadFrontend)
+    case JsDefined(unknown) =>
+      Left(UnexpectedState(s"Unsupported file upload service '$unknown'. Only 'upscan' or 'fileUpload' are supported"))
+    case JsUndefined() => Right(FileUploadProvider.FileUploadFrontend)
+  }
+
```
girishkamat, line 72: Minor: we can avoid usage of  JsDefine, JsString etc

```
  lazy val optFileUploadProvider1 = (json \ "service").asOpt[String].fold[Opt[FileUploadProvider]](Right(FileUploadProvider.FileUploadFrontend)) {
    case "upscan" => Right(FileUploadProvider.Upscan)
    case "fileUpload" => Right(FileUploadProvider.FileUploadFrontend)
    case unknown => Left(UnexpectedState(s"Unsupported file upload service '$unknown'. Only 'upscan' or 'fileUpload' are supported"))
  }
```

josef-vlach: This definition is not the same and when `(json \ "service")` is not a String it pass, instead of warning you.

girishkamat (approved):

PR #688, opened 2021-09-23
GFORMS-1418 Output summary & acknowledgement pdfs using Gform pdf gen…

girishkamat:
…erator


josef-vlach:
Shouldn't we now remove this check:  https://github.com/hmrc/gform/blob/master/app/uk/gov/hmrc/gform/form/FormService.scala#L125-L128

girishkamat:
> Shouldn't we now remove this check: https://github.com/hmrc/gform/blob/master/app/uk/gov/hmrc/gform/form/FormService.scala#L125-L128

No longer needed i guess. But we still have printPDF and notification PDF that rely on PDF generator. Do you know where they are used?

josef-vlach:
printPDF and notificationPDF should be converted as well, so we can move away from PDF generator microservice completely.

For usage  look at `"printSection"` in our templates.

josef-vlach (approved):
Cannot we now remove that

PR #690, opened 2021-10-08
GFORMS-1445 - Upload file to upscan ie. integration with upscan-initiate

josef-vlach:


app/uk/gov/hmrc/gform/upscan/UpscanService.scala
```diff
@@ -0,0 +1,49 @@
...
+
+class UpscanService(
+  upscanConnector: UpscanConnector,
+  upscanRepository: UpscanRepository
+)(implicit ec: ExecutionContext)
+    extends UpscanAlgebra[Future] {
+
+  def download(
+    downloadUrl: String
+  ): Future[ByteString] = upscanConnector.download(downloadUrl)
```
rhubner, line 31: What is the maximum size of file? This can blow up whole service.

app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -100,6 +101,10 @@ class ApplicationModule(context: Context)
 
   val formMetadaModule = new FormMetadataModule(mongoModule)
 
+  val queryParameterCrypto =
```
rhubner, line 104: Is there any reason why we creating this manually and not using `uk.gov.hmrc.crypto.ApplicationCrypto` from `hmrc-crypto`?

app/uk/gov/hmrc/gform/upscan/UpscanService.scala
josef-vlach: Good observation. Any limit at this moment is imposed by upscan only.

There are two reasons why this is done.

 1. This implementation is prototype only, not everything is mean to stay for long.
 2. Following ticket is about taking this image and compressing it. And as far as I know there is no streaming compression solution for JVM, all implementations I've found are `Array[Byte] => Array[Byte]`. 
 
 To elaborate on point 1., ultimate goal is to spin up a new service which will be doing transferring/compressing of images and uploading them to `file-upload`

PR #691, opened 2021-10-12
GFORMS-1492 - Introduce a new property to store old formId(s) of form to retrieve saved data

DursunGundogan:


conf/app.routes
```diff
@@ -6,6 +6,8 @@ POST          /new-form/:formTemplateId/:userId/:affinityGroup
 PUT           /forms/:userId/:formTemplateId                                                            uk.gov.hmrc.gform.form.FormController.updateFormDataPlain(userId: UserId, formTemplateId: String)
 PUT           /forms/:userId/:formTemplateId/:accessCode                                                uk.gov.hmrc.gform.form.FormController.updateFormData(userId: UserId, formTemplateId: String, accessCode: AccessCode)
 
+POST          /forms/:userId/:formTemplateId/from-legacy                                                uk.gov.hmrc.gform.form.FormController.createFormFromLegacy(userId: UserId, formTemplateId: String)
```
josef-vlach, line 9: Let's use `FormTemplateId` type instead of `String`. (Other places where `String` is used for `formTemplateId`, are probably overlooks)

app/uk/gov/hmrc/gform/form/FormService.scala
```diff
@@ -142,6 +142,19 @@ class FormService[F[_]: Monad](
...
+  def createFormFromLegacy(formIdData: FormIdData, newFormIdData: FormIdData)(implicit hc: HeaderCarrier): F[Unit] = {
+    val lowerCased = newFormIdData.lowerCaseId
+
+    for {
+      form <- get(formIdData)
+      newForm = form.copy(_id = newFormIdData.toFormId, formTemplateId = newFormIdData.formTemplateId)
+      _ <- formPersistence.upsert(newForm)
+      _ <- formMetadataAlgebra.upsert(lowerCased)
+      _ <- formPersistence.delete(formIdData.toFormId)
+      - <- formMetadataAlgebra.delete(formIdData)
```
josef-vlach, line 154: What's going on with that minus sign `-` on this line?

conf/app.routes
josef-vlach: There is not endpoint for 
```
s"$baseUrl/forms/${userId.value}/${formTemplateId.value}/${accessCode.value}/from-legacy"
```
Not we need to ask if agent support is needed or not for this functionality.

PR #693, opened 2021-10-13
GFORMS-1492 - Introduce a new property to store old formId(s) of form to retrieve saved data

DursunGundogan:


app/uk/gov/hmrc/gform/formmetadata/FormMetadataService.scala
```diff
@@ -117,4 +118,12 @@ class FormMetadataService(formMetadataRepo: Repo[FormMetadata])(implicit ec: Exe
         case Left(UnexpectedState(error)) => Future.failed(new Exception(error))
         case Right(unit)                  => Future.successful(unit)
       }
+
+  override def delete(formIdData: FormIdData): Future[Unit] = {
+    val metadata = newFormMetadata(formIdData)
+    toFuture(formMetadataRepo.delete(metadata.formIdData.toFormId.value)) >>
```
josef-vlach, line 124: There exists Functorial version of Monadic `>>` operation. It is called `as`.

Add `import cats.syntax.functor._` and you can do just the logging without need to wrap it in the `Future`

josef-vlach (changes requested):
Looks good, just one small possible improvement.

josef-vlach (approved):

PR #713, opened 2021-12-08
Support for JDK 17

josef-vlach:


DursunGundogan (approved):

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.5.5
```
rhubner, line 1: Is this safe? Using RC1?

josef-vlach: It works for me on all projects I need to touch.

> If no serious issues are found by Saturday, December 11th 2021, 1.6.0-RC1 will become 1.6.0 final.

[Source](https://eed3si9n.com/sbt-1.6.0-beta)

PR #721, opened 2022-01-11
Parser library replacement

rhubner:


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -284,10 +284,10 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
     |
     quotedConstant)
 
-  lazy val quotedConstant: Parser[Expr] = "'" ~> anyConstant <~ "'" | "''".r ^^^ Constant("")
+  lazy val quotedConstant: Parser[Expr] = anyConstant | "''".r ^^^ Constant("")
 
-  lazy val anyConstant: Parser[Constant] = """[^']+""".r ^^ { str =>
-    Constant(str)
+  lazy val anyConstant: Parser[Constant] = """'[^']+'""".r ^^ { str =>
+    Constant(str.replaceAll("'", ""))
```
josef-vlach, line 290: Why is is replaceAll needed? Isn't `[^']+` regex saying match anything except `'` ?

rhubner: I need to make some changes to how I match this, because parser by default skip all whitespace. So I need to capture this including a single quotation mark. But I agree that we don't need `.replaceAll` and `substring` will be enough. I will make changes.

josef-vlach (approved):

PR #725, opened 2022-02-02
GFORMS-1521 - Store non-sensitive fields in plain text

DursunGundogan:


app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
```diff
@@ -52,38 +52,29 @@ object VisitIndex {
...
   formTemplateId: FormTemplateId,
   formTemplateVersion: Option[FormTemplateVersion],
-  formData: FormData,
   status: FormStatus,
   visitsIndex: VisitIndex,
-  thirdPartyData: ThirdPartyData,
   envelopeExpiryDate: Option[EnvelopeExpiryDate],
-  componentIdToFileId: FormComponentIdToFileIdMapping
+  componentIdToFileId: FormComponentIdToFileIdMapping,
+  sensitive: Sensitive
```
josef-vlach, line 61: I think this `sensitive` field was not a good idea after all. Would you mind to getting rid of this `sensitive: Sensitive` field and restore `userId`, `formData` and `thirdPartyData` fields back and encrypt them separately?

Motivation: It would be good to keep `Form` class in sync with frontend, because it would allow us to derive `Format[Form]` once we have all users migrated to this new `EncryptedFormFormat`. After that we would be able do define

```
object Form {
  implicit val format: OFormat[Form] = derived.oformat()
}
```
For interchange between frontend and backend.

josef-vlach (changes requested):

josef-vlach (approved):

PR #728, opened 2022-02-10
GFORMS-1657 - Add new component 'taxPeriodDate' to allow a date without a day

rhubner:


josef-vlach (changes requested):
Can you reword commit message 

GFORMS-1657 - Add new component 'taxPeriodDate' to allow a date without a day

josef-vlach:
That is not true. By default if PR has one commit squash will using commit message, not a PR title.

But you can edit the message as needed, so please do that.

josef-vlach (approved):

PR #732, opened 2022-02-18
GFORMS-1532 - gForm Admin -Add a saved data count per form in the gForm admin tool

DursunGundogan:


josef-vlach (changes requested):
It looks ok, only issue I have is with naming thing. 

Names `DataModule`, `DataController` etc. are too generic, providing no indication whatsoever about functionality those relates classes provide/relates to. 

Would it be possible to use `FormStatistics` instead of `Data`?

josef-vlach (approved):

PR #747, opened 2022-03-23
GFORMS-1712 - An invalid XML character (Unicode: 0x2) was found in th…

DursunGundogan:
…e element content of the document.Can't load the XML resource (using TRaX transformer). org.xml.sax.SAXParseException

- Remove unprintable chars from HTML


josef-vlach (changes requested):
This is not right solution. We don't want this character live in data anywhere.

Please add it to `INVISIBLE_UNICODE_DESC_MAPPING` instead.

And it would be nice to add other ctrl+* characters, not only single one https://unicodelookup.com/#ctrl (some of them we want to accept though)

PR #768, opened 2022-04-22
GFORMS-1722: configure width of summary list section on CYA

lxol:


josef-vlach (commented):
Why are these change needed?

lxol:
> 

the change to `SummarySection` format is needed to support a default value of `displayWidth`. Without that play-json can't parse the json of SummarySection with missing displayWidth

lxol:
the change to `displayWidth` Reads is to support "l" "m" "xl" DisplayWidth enum. I uppercase it as it is read automatically if the string is equal to the enumeration name. And they are all in capital cases.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/SummarySection.scala
```diff
@@ -25,7 +25,8 @@ case class SummarySection(
   header: SmartString,
   footer: SmartString,
   continueLabel: Option[SmartString],
-  fields: Option[NonEmptyList[FormComponent]]
+  fields: Option[NonEmptyList[FormComponent]],
+  displayWidth: DisplayWidth.DisplayWidth = DisplayWidth.M
```
josef-vlach, line 29: Ok, can we make this other type? This doesn't represent same information as `DisplayWidth.DisplayWidth`. Also we should accept only value which have actually some effect. We are now silently doing nothing if user say for example `xs` etc.

lxol: ok, makes sense. I will create a new type

lxol: I introduced `SummaryDisplayWidth` type in both `gform` and `gform-frontend`

josef-vlach (approved):

PR #779, opened 2022-05-12
GFORMS-1746: submission time is UTC time rather than local

lxol:


lxol:
The LocalDateTime.now returns time in UTC rather than in "Europe/London" timezone. 
The fix is to explicitly set it up in that timezone.
@josef-vlach I noticed that we use LocalDateTime.now and LocalDate.now in several places. I guess we have the same issue there as well. Shall we apply the same fix everywhere?

josef-vlach (changes requested):
The idea was to keep using UTC, with the exception of PDF (and that new xml field you've added recently)

lxol:
> The idea was to keep using UTC, with the exception of PDF (and that new xml field you've added recently)
if we use LocalDateTime.now we rely on the timezone that is set up on the server. In local environment `LocalDateTime.now` returns time in "Europe/London"  but on the server it is UTC. 
If we want UTC regardless of the environment shall we use `Instant.now` instead?

josef-vlach:
Hm, so we wanted to use `LocalDateTime`, but server runs in UTC, so it was never behaving as expected.

What would it mean for us to switch to `Europe/London` timezone wrt the whole functionality?

(I think to have `LocalDateTime` everywhere may be better option than to stick to UTC and do the conversion to BST on couple of places)

lxol:
> Hm, so we wanted to use `LocalDateTime`, but server runs in UTC, so it was never behaving as expected.
> 
> What would it mean for us to switch to `Europe/London` timezone wrt the whole functionality?
> 
> (I think to have `LocalDateTime` everywhere may be better option than to stick to UTC and do the conversion to BST on couple of places)

ok, I will leave it as it is but fix only the places that affects the issues mentioned in the ticket then

app/uk/gov/hmrc/gform/submission/handlebars/HandlebarsTemplateProcessorHelpers.scala
```diff
@@ -421,10 +421,10 @@ class HandlebarsTemplateProcessorHelpers(
   def isNotSuccessCode(code: Integer): CharSequence = log("isNotSuccessCode", code)(condition(!isSuccCode(code)))
   private def isSuccCode(code: Integer): Boolean = code != null && code >= 200 && code <= 299
 
-  def desCurrentDate: CharSequence = condition(DateTimeFormatter.ISO_LOCAL_DATE.format(timeProvider.localDateTime))
-  def currentDate: CharSequence = condition(DateTimeFormatter.BASIC_ISO_DATE.format(timeProvider.localDateTime))
+  def desCurrentDate: CharSequence = condition(DateTimeFormatter.ISO_LOCAL_DATE.format(timeProvider.localDateTime()))
+  def currentDate: CharSequence = condition(DateTimeFormatter.BASIC_ISO_DATE.format(timeProvider.localDateTime()))
```
josef-vlach, line 425: Now the obvious question is why are these stuck in UTC, do we expect that?

lxol: I thought that by sticking to UTC we make the server's and local environment consistent. So that localDateTime by default provides the local time on the server.
And only in some cases we can use the same function to get local date time in other time zones. We use it only once for `submittedDate`

josef-vlach: But the question is: Do we want for handlebar templates to stay stuck in UTC or should they produce BST dates?

The fact that implementation is using `timeProvider.localDateTime` I tend to believe that the intention was to use BST and not to stick to UTC.

josef-vlach: Ok, let's go back to to initial implementation. We should also modify all `LocalDate.now` and `LocalDateTime.now` accordingly to answer your initial question.

josef-vlach (approved):

PR #780, opened 2022-05-12
GFORMS-1134 - Upscan: Introduce a property to define what file types …

josef-vlach:
…are allowed in a form.


josef-vlach:
We have been mixing upscan errors with our validation errors leading to some mess in the code. This is attempt to keep track of where the upscan error originates from.

app/uk/gov/hmrc/gform/upscan/UpscanController.scala
```diff
@@ -108,18 +108,18 @@ class UpscanController(
               s"Upscan callback successful, fcId: $formComponentId, reference: ${upscanCallbackSuccess.reference}, fileMimeType: ${upscanCallbackSuccess.uploadDetails.fileMimeType}, fileName: ${upscanCallbackSuccess.uploadDetails.fileName}, size: ${upscanCallbackSuccess.uploadDetails.size}"
             )
             val allowedFileTypes: AllowedFileTypes = formTemplate.allowedFileTypes
-            val validated: Validated[FailureDetails, Unit] =
+            val validated: Validated[UpscanValidationFailure, Unit] =
               validateFile(allowedFileTypes, upscanCallbackSuccess.uploadDetails)
 
+            println("HeRe 232323")
+            println("validated: " + validated)
```
DursunGundogan, line 115: Do we need these `println()`s?

DursunGundogan (changes requested):

app/uk/gov/hmrc/gform/upscan/UpscanController.scala
josef-vlach: Thanks, removed.

PR #783, opened 2022-05-30
GFORMS-1744 - Allow choices to be displayed based on an includeIf exp…

DursunGundogan:


josef-vlach (changes requested):
This is missing integration with top level `"booleanExpressions"`

josef-vlach:
It also needs to rewrite choice related expressions using `=` to use `contains` instead, as is happening elsewhere.

josef-vlach (approved):

PR #789, opened 2022-07-13
Task List

josef-vlach:


app/uk/gov/hmrc/gform/formtemplate/FormTemplatesControllerRequestHandler.scala
```diff
@@ -101,6 +101,12 @@ object FormTemplatesControllerRequestHandler {
   private def list[A <: JsValue](reads: Reads[A]): Reads[JsArray] =
     Reads.list(reads).map(JsArray.apply)
 
+  // private def debug[A <: JsValue](r: Reads[A]): Reads[A] = Reads { json =>
```
lxol, line 104: is it still necessary?

lxol (approved):

app/uk/gov/hmrc/gform/formtemplate/FormTemplatesControllerRequestHandler.scala
josef-vlach: No, but I find this very useful anytime we are pre-processing the json and something not working as expected. 
I'm tired of always having to write it from scratch, so I thought why not to leave it in as a comment.

PR #793, opened 2022-07-26
GFORMS-1835: expression capability for country lookup column values

lxol:


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -210,6 +210,9 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
       case _ ~ _ ~ dataRetrieveId ~ _ ~ dataRetrieveAttribute =>
         DataRetrieveCtx(DataRetrieveId(dataRetrieveId), DataRetrieveAttribute.fromName(dataRetrieveAttribute))
     } // to parse date form fields with offset or date constants i.e TODAY, 01012020 etc (with or without offset)
+    | FormComponentId.unanchoredIdValidation ~ ".column." ~ alphabeticOnly ^^ { case value ~ _ ~ column =>
+      CsvCountryCheck(FormComponentId(value), column)
```
josef-vlach, line 214: When we talked about safety, my understanding that we want to prevent people to upload invalid column name, by validation `alphabeticOnly` for allowed names?

josef-vlach: Actually I'm taking this back. This is ok, we should do the check in runtime and hard fail if we can't find the column name.

josef-vlach (approved):

PR #794, opened 2022-07-29
GFORMS-1831: Add support so that headers can be included in miniSummaryList

lxol:


app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
```diff
@@ -170,18 +170,30 @@ class FormComponentMaker(json: JsValue) {
...
       jsValues <- toOpt((json \ "rows").validate[List[JsValue]], "/rows")
-      rs       <- jsValues.map(getRow).sequence
+      rs <- jsValues
+              .map(j =>
+                getValueRow(j) match {
+                  case Right(r) => Right(r)
+                  case _        => getHeaderRow(j)
+                }
+              )
+              .sequence
```
VlachJosef, line 196: `.map` followed by `.sequence` is what what `.traverse` do in one step, so please use `.traverse`

VlachJosef (changes requested):

app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
josef-vlach, line 196: `.map` followed by `.sequence` is what what `.traverse` do in one step, so please use `.traverse`

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/formtemplate/FormComponentMaker.scala
lxol: done

josef-vlach (approved):

PR #799, opened 2022-08-08
GFORMS-1837 -Make the last section and task always be the submit task…

DursunGundogan:
… instead of a button


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/FormTemplate.scala
```diff
@@ -43,6 +43,7 @@ case class FormTemplate(
   languages: AvailableLanguages,
   save4LaterInfoText: Option[Save4LaterInfoText],
   summarySection: SummarySection,
+  submitSection: Option[SubmitSection],
```
josef-vlach, line 46: As I've said in the frontend review. I think this should be mandatory field.

josef-vlach (changes requested):

josef-vlach (approved):

PR #856, opened 2022-12-01
GFORMS-1989 - Deprecate limit property in addToList and replace with …

DursunGundogan:
…repeatsUntil and repeatsWhile


app/uk/gov/hmrc/gform/formtemplate/Rewriter.scala
```diff
@@ -162,9 +162,29 @@ trait Rewriter {
...
+            _,
+            _,
+            _,
+            _,
+            fields,
+            _
+          ) =>
         includeIf.toList ++ pages.toList.flatMap(_.includeIf.toList) ++ fields.fold(List.empty[IncludeIf])(
-          _.toList.flatMap(_.includeIf.toList)
+          _.toList.flatMap(_.includeIf.toList) ++ repeatsUntil.toList ++ repeatsWhile.toList
```
josef-vlach, line 187: This should not assume there must be `fields` field.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/formtemplate/Rewriter.scala
DursunGundogan: Fixed

josef-vlach (approved):

PR #861, opened 2022-12-15
GFORMS-1974 - Add ability to hide the Save and come back later button…

DursunGundogan:
… from a page


josef-vlach (changes requested):
According ticket `hideSaveAndComeBackButton` needs to be a section property, not a form property.

josef-vlach (approved):

PR #872, opened 2023-01-05
GFORMS-2019 - Include language used in dataOutput xml and json files

DursunGundogan:


josef-vlach (approved):
Looks good, but isn't there also frontend part? I don't see any PR for that.

PR #876, opened 2023-01-11
GFORMS-1515 - Add a property to output summary and acknowledgement PD…

DursunGundogan:
…Fs in tabular format

### DIFF
--- a/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
+++ b/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
@@ -129,8 +129,8 @@ case class AcknowledgementSection(
   shortName: Option[SmartString],
   fields: List[FormComponent],
   showReference: Boolean,
-  pdf: Option[SectionPdf],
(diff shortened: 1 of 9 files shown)


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/Section.scala
```diff
@@ -156,12 +156,16 @@ object AcknowledgementSection {
 
 }
 
-case class AcknowledgementSectionPdf(header: Option[SmartString], footer: Option[SmartString])
+case class SectionPdf(
```
josef-vlach, line 159: Why was this renamed? `SectionPdf` feels like a strange name, it is hard to guess what does it mean.

DursunGundogan: We should reuse it both case classes `AcknowledgementSection` and `SummarySection`. I'll try to find a better name for it.

josef-vlach (approved):

PR #924, opened 2023-03-24
GFORMS-2084 - Investigate how to use config to make adding calls to n…

josef-vlach:
…ew APIs easier to implement


lxol (approved):

DursunGundogan:
When I add an includeIf with a number result, it does not work. I think we should convert the result to Int for includeIfs
`"includeIf": "${dataRetrieve.ninoCheck.riskScore > 0}",`

DursunGundogan (approved):

PR #930, opened 2023-04-03
GFORMS-2136 - Add feature to save the content of the choice selected …

DursunGundogan:
…so that when choices change (i.e. new tax year) the users actual choice is saved

### DIFF
--- a/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
@@ -565,6 +565,10 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
         SelectionCriteriaSimpleValue(List(value))
       }
 
+  lazy val optionDataValue: Parser[OptionDataValue] = "TY_${" ~ _expr1 ~ "}" ^^ { case _ ~ expr ~ _ =>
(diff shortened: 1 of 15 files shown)


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -565,6 +565,10 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
         SelectionCriteriaSimpleValue(List(value))
       }
 
+  lazy val optionDataValue: Parser[OptionDataValue] = "TY_${" ~ _expr1 ~ "}" ^^ { case _ ~ expr ~ _ =>
```
josef-vlach, line 568: "TY_" is supposed to be a free text (except white spaces I guess)

josef-vlach (changes requested):

josef-vlach (changes requested):
Couple of comments, but also we should add validation for the following:
```
{
  "en": "6 April ${currentYear-1} to 5 April ${currentYear}",
  "value": "test_${currentYear}${currentYear}"
},
```
We should reject `value` if it contains expression(s) and it has not been matched by `optionDataValue: Parser[OptionDataValue]`

josef-vlach (approved):

PR #940, opened 2023-04-18
GFORMS-2152: Add a new property in destinations that calls an API sev…

lxol:
…eral times for a payload in an array

### DIFF
--- a/app/uk/gov/hmrc/gform/submission/handlebars/HandlebarsHttpApiSubmitter.scala
+++ b/app/uk/gov/hmrc/gform/submission/handlebars/HandlebarsHttpApiSubmitter.scala
@@ -50,8 +50,8 @@ class RealHandlebarsHttpApiSubmitter[F[_]](
     def parseAndProcessAsList(input: String): List[String] =
       Try(Json.parse(input)).toOption
         .flatMap {
-          case jsonArray: JsArray if destination.multiRequestPayload => Some(jsonArray.value.map(_.toString).toList)
-          case _                                                     => None
+          case jsonArray: JsArray => Some(jsonArray.value.map(_.toString).toList)
+          case _                  => None
         }
         .getOrElse(List(input))
         .map {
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/gform/submission/handlebars/HandlebarsHttpApiSubmitter.scala
```diff
@@ -52,30 +73,62 @@ class RealHandlebarsHttpApiSubmitter[F[_]](
...
-              )
-            }
-            httpClient.put(uri, body)
+            val bodies = destination.payload.fold(List(""))(body => parseAndProcessAsList(body))
+            callUntilError(
+              bodies.map((uri, _)),
+              httpClient.put,
+              modelTree.value.formId,
+              destination.id
+            )
```
josef-vlach, line 94: Can we keep handling of `destination.multiRequestPayload` handling login separated from current single payload implementation? This implementation is hard to follow.

Simple `if .. else` should do the trick
```
if (destination.multiRequestPayload) {
   // call parseAndProcessAsList with the assumption it is only for multiRequestPayload
   // Feel free to put code for multiRequestPayload into its own class etc.
} else {

  destination.method match {
    case HttpMethod.GET => httpClient.get(uri)
    case HttpMethod.POST => ... current code unchanged
}
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/submission/handlebars/HandlebarsHttpApiSubmitter.scala
lxol: refactored to keep the current code unchanged when multiRequestPayload = false

josef-vlach (commented):
I think there is quite a space for an improvement. My quick attempt is here : https://github.com/hmrc/gform/pull/940/commits/854abd58b1c78aa7096959d05fbefa79f7771789
If you are ok with the changes, I'll approve this.

lxol:
> 

I simplified it even further. Now 3 arguments of `callUntilError` can be acquired from the scope

josef-vlach (approved):

PR #945, opened 2023-04-25
GFORMS-2144 -Only display character count message after a user has en…

DursunGundogan:
…tered a certain amount of chars


app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala
```diff
@@ -1294,6 +1294,25 @@ object FormTemplateValidator {
...
+  def validateDataThreshold(sectionsList: List[Page]): ValidationResult = {
+    def check(textArea: TextArea): Boolean = textArea match {
+      case TextArea(_, _, _, _, _, Some(value)) if value > 100 => true
+      case _                                                   => false
+    }
+
+    val textAreaFieldIds = allFormComponents(sectionsList)
+      .map(fv => (fv.id, fv.`type`))
+      .collect {
+        case (fId, textArea: TextArea) if check(textArea) =>
```
josef-vlach, line 1307: You should be able to replace this with `IsTextArea`

josef-vlach (changes requested):

PR #954, opened 2023-05-10
GFORMS-2149 - Move handlebar payload configuration to a separate conf…

DursunGundogan:
…ig file

### DIFF
--- a/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/destinations/Destination.scala
+++ b/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/destinations/Destination.scala
@@ -32,9 +32,6 @@ import uk.gov.hmrc.gform.sharedmodel.formtemplate.destinations.DestinationInclud
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.destinations.Destination.SubmissionConsolidator
 import uk.gov.hmrc.gform.sharedmodel.notifier.NotifierPersonalisationFieldId
 
-import java.nio.file.Paths
-import scala.io.Source
-
 sealed trait DestinationWithCustomerId {
   def customerId(): Expr
 }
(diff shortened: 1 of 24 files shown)


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/destinations/Destination.scala
```diff
@@ -411,4 +423,28 @@ object UploadableConditioning {
...
+    val path = Paths.get(url).normalize().toAbsolutePath().toString()
+    val source = Source.fromFile(path)
+    try Some(source.mkString)
+    catch {
+      case ex: Exception =>
+        throw new RuntimeException(
+          s"An error occurs when attempting to read the file from $path in the ${id.id} destination with exception : $ex"
+        )
+    } finally source.close()
+  }
```
josef-vlach, line 441: What is the idea behind this? I'm not sure what is this for?

DursunGundogan: The idea is to read and load the handlebar configuration into the `payload` attribute while posting a form template if payloadUrl is present.

josef-vlach (changes requested):
**External payload needs to be *inlined* into template on upload time....**

When template is uploaded and corresponding external payload do not exists, we need to raise an error. 
Also when external payload is being uploaded, we need to update corresponding formTemplate inline version of it (at the beginning it may not exists.)
If it comes to the question what needs to be first payload or template, the answer is payload.

One indication that this was done right is that gform-frontend would contains exactly 0 changes.

Notes. This only applies for `"type": "handlebarsHttpApi"` destination when `"payload"` property is not defined.

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/destinations/Destination.scala
```diff
@@ -81,10 +79,12 @@ object Destination {
     uri: String,
     method: HttpMethod,
     payload: Option[String],
+    payloadName: Option[String],
```
josef-vlach, line 82: `payloadName` should be derived. It should equal to `formTemplateId-destinationId`

josef-vlach (changes requested):

josef-vlach (approved):

PR #966, opened 2023-06-09
GFORMS-2208 - Prevent forms being posted with a booleanExpression tha…

DursunGundogan:
…t references itself


josef-vlach (commented):
There is `TopLevelExpressions` which is able to detect cycles (in `expressions`, I'm not sure about `booleanExpressions`), it doesn't detect self references though, but that should be easy to add. Can you you investigate that direction? It is not ideal to have scattered similar logic in various places.

josef-vlach (approved):

PR #986, opened 2023-07-11

DursunGundogan:
…data store


app/uk/gov/hmrc/gform/submission/destinations/DestinationSubmissionInfo.scala
```diff
@@ -17,8 +17,13 @@
...
 
 import uk.gov.hmrc.gform.sharedmodel.form.FormId
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.FormTemplateId
 import uk.gov.hmrc.gform.submission.Submission
 
-case class DestinationSubmissionInfo(customerId: String, submission: Submission) {
+case class DestinationSubmissionInfo(
+  customerId: String,
+  submission: Submission,
+  formTemplateId: FormTemplateId
```
lxol, line 26: submission.dmsMetaData.formTemplateId 
is that the same formTemplateId as a new `formTemplateId` argument?

DursunGundogan: That's the same formTemplateId to pass into DestinationSubmitter.

DursunGundogan: FormTemplateId exists in DmsMetaData. I'll remove it.

app/uk/gov/hmrc/gform/submission/DataStoreFileGenerator.scala
```diff
@@ -0,0 +1,60 @@
...
+    val userSessionJson: Json = Json.obj(
+      "clientIp"      -> Json.fromString(userSession.clientIp),
+      "deviceId"      -> Json.fromString(userSession.deviceId),
+      "userAgent"     -> Json.fromString(userSession.userAgent),
+      "relativePath"  -> Json.fromString(userSession.relativePath),
+      "credentialId"  -> Json.fromString(userSession.credentialId),
+      "affinityGroup" -> Json.fromString(userSession.affinityGroup.map(_.toString).getOrElse("")),
+      "authEmail"     -> Json.fromString(userSession.authEmail),
+      "authPhone"     -> Json.fromString(userSession.authPhone),
+      "enrolments"    -> Json.arr(userSession.enrolments.map(playJsonToCirce): _*)
```
lxol, line 50: looks like enforlments from userSession is only used as circe Json never as play.json JsValue. Would it be good idea to format it like that in the first place
```
import io.circe.Json
import play.api.libs.json._
import play.api.libs.functional.syntax._

case class UserSession(
  clientIp: String,
  deviceId: String,
  userAgent: String,
  relativePath: String,
  credentialId: String,
  affinityGroup: Option[AffinityGroup],
  authEmail: String,
  authPhone: String,
  enrolments: List[Json]
)

object UserSession {
  val empty = UserSession("", "", "", "", "", None, "", "", Nil)

  implicit val enrolmentsReads: Reads[List[Json]] = (JsPath \ "enrolments").read[List[JsValue]].map(_.map(playJsonToCirce).collect {
    case Right(json) => json
  })

  implicit val enrolmentsWrites: Writes[List[Json]] = (JsPath \ "enrolments").write[List[JsValue]].contramap(_.map(jsonToPlayJson))

  implicit val format: OFormat[UserSession] = Json.format[UserSession]

  private def playJsonToCirce(jsValue: JsValue): Either[String, Json] = ???
  private def jsonToPlayJson(json: Json): JsValue = ???
}
something like that....

File: app/uk/gov/hmrc/gform/submission/DataStoreFileGenerator.scala [hunk 3]
```diff
@@ -16,45 +16,50 @@
...
+      val userSessionJson: JsObject = Json.obj(
+        "clientIp"      -> JsString(userSession.clientIp),
+        "deviceId"      -> JsString(userSession.deviceId),
+        "userAgent"     -> JsString(userSession.userAgent),
+        "relativePath"  -> JsString(userSession.relativePath),
+        "credentialId"  -> JsString(userSession.credentialId),
+        "affinityGroup" -> JsString(userSession.affinityGroup.map(_.toString).getOrElse("")),
+        "authEmail"     -> JsString(userSession.authEmail),
+        "authPhone"     -> JsString(userSession.authPhone),
+        "enrolments"    -> JsArray(userSession.enrolments.map(Json.toJson(_)))
```
lxol, line 52: enrolement is already JsValue. We don't need to to convert it again.
```
        "enrolments"    -> JsArray(userSession.enrolments)
```

lxol (approved):

PR #987, opened 2023-07-20

DursunGundogan:
…data store


app/uk/gov/hmrc/gform/sdes/SdesConnector.scala
```diff
@@ -33,7 +33,7 @@ class SdesConnector(wSHttp: WSHttp, sdesBaseUrl: String, sdesBasePath: String, h
     HttpReadsInstances.throwOnFailure(HttpReadsInstances.readEitherOf(HttpReadsInstances.readRaw))
 
   def notifySDES(payload: SdesNotifyRequest)(implicit hc: HeaderCarrier): Future[HttpResponse] = {
-    logger.debug(s"SDES notification request: ${Json.stringify(Json.toJson(payload))}")
+    logger.info(s"SDES notification request: ${Json.stringify(Json.toJson(payload))}")
```
lxol, line 36: does SdesNotifiyRequest contains personal information? We can see `info` logs in prod kibana.

DursunGundogan: It does not contain personal information. We should see the notification logs in kibana.

lxol: It looks like it contains a file name. Is that any name ? Can it be like "JohnDoe_TaxReturns_2023.pdf" ?

DursunGundogan: No, it looks like ``envelopeId`.zip`. The file name does not contain any personal information.

lxol (approved):

PR #989, opened 2023-07-24

DursunGundogan:
…data store

### DIFF
--- a/app/uk/gov/hmrc/gform/time/TimeProvider.scala
+++ b/app/uk/gov/hmrc/gform/time/TimeProvider.scala
@@ -20,7 +20,7 @@ import java.time._
 class TimeProvider {
   def localDateTime(): LocalDateTime = LocalDateTime.now(ZoneId.of("Europe/London"))
   def instant(): Instant = Instant.now
-  def instantLocal(): Instant = localDateTime().toInstant(ZoneOffset.UTC)
+  def instantLocal(): Instant = LocalDateTime.now().atZone(ZoneId.of("Europe/London")).toInstant
 }
 
 class TimeModule {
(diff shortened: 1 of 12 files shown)


app/uk/gov/hmrc/gform/time/TimeProvider.scala
```diff
@@ -20,6 +20,7 @@ import java.time._
 class TimeProvider {
   def localDateTime(): LocalDateTime = LocalDateTime.now(ZoneId.of("Europe/London"))
   def instant(): Instant = Instant.now
+  def instantLocal(): Instant = localDateTime().toInstant(ZoneOffset.UTC)
```
lxol, line 23: I think you should respect timezone of the localdatetime, otherwise when we transition to/from summer time there will be 1 hour gap. Here is an example for transition on 26 Marh this year
```
val dstStart = LocalDateTime.of(2023, 3, 26, 0, 59)  // DST starts on March 26, 2023 at 1:00 in London
// dstStart: LocalDateTime = 2023-03-26T00:59

// Get LocalDateTime for one minute before and after the DST transition
val oneMinuteBefore = dstStart
// oneMinuteBefore: LocalDateTime = 2023-03-26T00:59
val oneMinuteAfter = dstStart.plusHours(1).plusMinutes(2)  // add 1 hour because of the DST transition
// oneMinuteAfter: LocalDateTime = 2023-03-26T02:01


oneMinuteBefore.toInstant(ZoneOffset.UTC)
// res1: Instant = 2023-03-26T00:59:00Z
oneMinuteAfter.toInstant(ZoneOffset.UTC)
// res2: Instant = 2023-03-26T02:01:00Z


val zone = ZoneId.of("Europe/London")
// zone: ZoneId = Europe/London
oneMinuteBefore.atZone(zone).toInstant
// res3: Instant = 2023-03-26T00:59:00Z
oneMinuteAfter.atZone(zone).toInstant
// res4: Instant = 2023-03-26T01:01:00Z

```
in  UTC zone there is a 1 hour gap

PR #1008, opened 2023-09-14
GFORMS-2346 - Make confirmation property work in task list forms

DursunGundogan:


app/uk/gov/hmrc/gform/formtemplate/FormTemplatesControllerRequestHandler.scala
```diff
@@ -345,7 +345,12 @@ object FormTemplatesControllerRequestHandler {
           )
         })
 
-      json.transform(updatePageQuestions andThen updateSectionQuestions) orElse JsSuccess(json)
+      val updateTaskList =
+        (__ \ "sections").json.update(list((__ \ "tasks").json.update(list(updateSectionQuestions))))
+
+      json.transform(
+        updateTaskList andThen updatePageQuestions andThen updateSectionQuestions
```
josef-vlach, line 352: This also needs to cover cases when confirmation question is in ATL in task list

josef-vlach (changes requested):

PR #1017, opened 2023-10-03
GFORMS-2367 - Convert the version from a string to a number

josef-vlach:


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/destinations/Destination.scala
```diff
@@ -248,7 +248,7 @@ case class UploadableDataStoreDestination(
   includeIf: DestinationIncludeIf,
   failOnError: Option[Boolean],
   formId: FormId,
-  version: FormTemplateVersion,
```
lxol, line 251: Why you relaxed the specific type to the string?

josef-vlach: I'm not sure what this field is for, but it is not a `FormTemplateVersion`

lxol (approved):

PR #1020, opened 2023-10-09
GFORMS-2339: Add a checkbox to keep section/page title aligned with f…

lxol:
…ield label

### DIFF
--- a/app/uk/gov/hmrc/gform/builder/BuilderController.scala
+++ b/app/uk/gov/hmrc/gform/builder/BuilderController.scala
@@ -31,6 +31,7 @@ import uk.gov.hmrc.gform.core._
 import uk.gov.hmrc.gform.formtemplate.{ FormTemplateService, FormTemplatesControllerRequestHandler, RequestHandlerAlg }
 import uk.gov.hmrc.gform.history.HistoryService
 import uk.gov.hmrc.gform.sharedmodel.formtemplate._
+import io.circe.generic.semiauto._
 
 import scala.concurrent.ExecutionContext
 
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -380,15 +382,91 @@ class BuilderController(
...
+    final def apply(cursor: HCursor): Decoder.Result[ComponentDetails] =
+      cursor.get[Json]("componentData").map(p => ComponentDetails(p.dropNullValues))
+  }
+
+  implicit val sectionUpdateDecoder: Decoder[SectionUpdateRequest] = new Decoder[SectionUpdateRequest] {
+    final def apply(c: HCursor): Decoder.Result[SectionUpdateRequest] = {
+      val payload = c.downField("componentDetails").delete.top.getOrElse(c.value)
+      val componentDetails = c.get[Option[ComponentDetails]]("componentDetails")
+      componentDetails.map(s => SectionUpdateRequest(payload, s))
+    }
```
josef-vlach, line 470: Would it be possible to derive these decoders?

lxol: We can make derive SectionDetails and ComponentDetails but not SectionUpdateRequest and ComponentUpdateRequest like here:
```
case class SectionDetails(sectionData: Json, sectionPath: String)

object SectionDetails {
  implicit val decodeSectionDetails: Decoder[SectionDetails] = deriveDecoder[SectionDetails]
}

case class ComponentDetails(componentData: Json)

object ComponentDetails {
  implicit val decodeComponentDetails: Decoder[ComponentDetails] = deriveDecoder[ComponentDetails]
}

case class ComponentUpdateRequest(payload: Json, sectionDetails: Option[SectionDetails])

case class SectionUpdateRequest(payload: Json, componentDetails: Option[ComponentDetails])

object Decoders {
  implicit val componentUpdateDecoder: Decoder[ComponentUpdateRequest] = new Decoder[ComponentUpdateRequest] {
    final def apply(c: HCursor): Decoder.Result[ComponentUpdateRequest] = {
      val payload = c.downField("sectionDetails").delete.top.getOrElse(c.value)
      val sectionDetails = c.get[Option[SectionDetails]]("sectionDetails")
      sectionDetails.map(s => ComponentUpdateRequest(payload, s))
    }
  }
  implicit val sectionUpdateDecoder: Decoder[SectionUpdateRequest] = new Decoder[SectionUpdateRequest] {
    final def apply(c: HCursor): Decoder.Result[SectionUpdateRequest] = {
      val payload = c.downField("componentDetails").delete.top.getOrElse(c.value)
      val componentDetails = c.get[Option[ComponentDetails]]("componentDetails")
      componentDetails.map(s => SectionUpdateRequest(payload, s))
    }
  }
}
```

lxol: However we can fully derive it if we completely change the current update requests payload to have optional fields for both section and component payloads

josef-vlach: Why both of them need to be optional?

lxol: One request is for section update an I modified a request to have optionally update component.
Anohter request is for component update and I modified it to optionally update section.
If we are about to merge then then both should be optional I think

josef-vlach (approved):

PR #1025, opened 2023-10-20
GFORMS-2321: When deploying a hbs payload template the form that uses it should be automatically reloaded

lxol:


lxol:
some rewiring's done to avoid circular dependencies

app/uk/gov/hmrc/gform/handlebarstemplate/HandlebarsTemplateController.scala
```diff
@@ -36,11 +45,24 @@ class HandlebarsTemplateController(
   def upsert(handlebarsTemplateId: HandlebarsTemplateId) = Action.async(parse.text) { implicit request =>
     val handleBarsTemplate = HandlebarsTemplate(handlebarsTemplateId, request.body)
 
-    handlebarsTemplateAlgebra.save(handleBarsTemplate).map { _ =>
-      NoContent
-    }
+    val formTemplateIdStr = handlebarsTemplateId.value.replaceAll("-(\\w+)$", "")
```
josef-vlach, line 48: What are the naming convention again? I wonder what is this doing?

lxol: it is here:
https://github.com/hmrc/gform/blob/95b742a38dcb119cfe15b8621978bf1564bf0524/app/uk/gov/hmrc/gform/formtemplate/FormTemplateService.scala#L145
`val templateId = HandlebarsTemplateId(s"${formTemplate._id.value}-${destId.id}")`
I assume that the destId.id doesn't have '-'. Otherwise we are in trouble.  We can come up with something more distinctive to separate templateId from the destId

PR #1036, opened 2023-11-01
GFORMS-2282: Automatically insert "languages": ["en","cy"] into form when inserting Welsh translations

lxol:


app/uk/gov/hmrc/gform/translation/TranslationController.scala
```diff
@@ -104,22 +90,32 @@ class TranslationController(
...
+        case (key, value) if key == "languages" => (key, Json.toJson(Seq("en", "cy")))
+        case otherwise                          => otherwise
+      }
+    } else {
+      val (before, after) = fields.span(_._1 != "version")
+      after match {
+        case version :: as => before ++ Seq(version, ("languages", Json.toJson(Seq("en", "cy")))) ++ as
+        case _             => before ++ Seq(("languages", Json.toJson(Seq("en", "cy"))))
+      }
+    }
```
josef-vlach, line 116: Can we use `===` and `=!=` instead of `==` and `!=` to make ensure we are comparing same types?

josef-vlach: Also it would be nice to define `Json.toJson(Seq("en", "cy"))` just once

lxol: done

josef-vlach (approved):

PR #1057, opened 2023-11-29
GFORMS-2378 - Improve SDES Alert logic

DursunGundogan:


app/uk/gov/hmrc/gform/sdes/SdesConnector.scala
```diff
@@ -38,5 +38,11 @@ class SdesConnector(wSHttp: WSHttp, sdesBaseUrl: String, sdesBasePath: String, h
     wSHttp
       .POST[SdesNotifyRequest, HttpResponse](url, payload, headers)
       .failWithNonSuccessStatusCodes(url)
+      .recoverWith { case e =>
+        logger.error(
+          s"Failed to notify SDES [file-ready] for the information type ${payload.informationType}, error : $e"
+        )
+        Future.failed(e)
+      }
```
josef-vlach, line 46: I think this is not right place to log. You are ignoring the exception [here](https://github.com/hmrc/gform/blob/main/app/uk/gov/hmrc/gform/scheduler/QueueAlgebra.scala#L36) and that is the place where we should be logging the error

josef-vlach (approved):

app/uk/gov/hmrc/gform/scheduler/QueueAlgebra.scala
```diff
@@ -26,14 +27,17 @@ trait QueueAlgebra[Q] {
...
 
   implicit val executionContext: ExecutionContext
 
   def sendWorkItem(workItem: WorkItem[Q]): Future[Unit]
 
   def processThenMarkAsComplete(acc: Seq[WorkItem[Q]], workItem: WorkItem[Q]): Future[Seq[WorkItem[Q]]] =
     sendWorkItem(workItem).map(_ => repo.completeAndDelete(workItem.id)).map(_ => acc :+ workItem).recoverWith {
-      case _ => repo.failed(workItem, maxFailureCount).map(_ => acc)
+      case e =>
+        logger.error(s"Failed to send the work-item : $e")
```
josef-vlach, line 39: I wonder if this shouldn't be rather this:

```suggestion
        logger.error(s"Failed to send the work-item", e)
```

josef-vlach (approved):

PR #1058, opened 2023-11-29
GFORMS-2444 - When outputting form data for the DataStore destination…

DursunGundogan:
… exclude DMS and info items


josef-vlach (changes requested):
I think there is some misunderstanding of objectives for this ticket.

The ticket is focused solely on `hmrcIlluminate` destination and json this destination produces. As such I can't see any fixes for said json as per GFORMS-2444

PR #1068, opened 2023-12-13
GFORMS-2473: Gform frontend - test-only endpoints should start with /…

lxol:
…test-only prefix


conf/testOnlyDoNotUseInAppConf.routes
```diff
@@ -11,7 +11,7 @@
 
 # Add all the application routes to the prod.routes file
 ->            /                                                                  prod.Routes
-->            /gform/builder                                                     builder.Routes
+->            /gform/test-only/builder                                           builder.Routes
```
josef-vlach, line 14: Why this change? platops-pr-bot has not been complaining about this.

PR #1077, opened 2024-01-05
GFORMS-1976: Add functionality to the toolbox link in test environmen…

lxol:
…ts to export and import form data from/to user session


app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -181,52 +182,56 @@ class ApplicationModule(context: Context)
...
-            Indexes.ascending("submitDetails.createdAt"),
-            IndexOptions()
-              .background(false)
-              .name("submittedIndex")
-              .expireAfter(submittedExpiry, TimeUnit.MILLISECONDS)
-          )
-        )
-        MongoUtils.ensureIndexes(this.collection, indexes, true)
-      }
-    },
```
josef-vlach, line 222: We should not be reusing this for snapshot data.

app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -272,13 +277,17 @@ class ApplicationModule(context: Context)
     )
   private val obligationModule = new ObligationModule(wSHttpModule, configModule)
   private val employmentsModule = new EmploymentsModule(wSHttpModule, configModule)
+
+  val savedFormsMongoCache = createMongoCacheRepository("savedForms")
```
josef-vlach, line 281: `savedForms` is overloaded name, let's call this `snapshots`

app/uk/gov/hmrc/gform/testonly/SavedForm.scala
```diff
@@ -0,0 +1,46 @@
...
+import play.api.libs.json.OFormat
+import play.api.libs.json.Reads._
+import play.api.libs.json._
+
+case class SavedForm(
+  templateId: String,
+  formId: String,
+  userId: String,
+  savedAt: Instant
+)
```
josef-vlach, line 31: Let call this `Snaphost`. 

Why this stores `formId` and `userId`?

It should store the raw data in something like `Map[String, String]`. Perhaps we can reuse `FormData`.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/ApplicationLoader.scala
lxol: We don't reuse `form` repository for snapshot data. It is just refactoring. Production code is not being changed. We just use the production instance of the `form` repository in the TestOnlyFormService to copy/restore to/from snapshot repository.

app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
```diff
@@ -0,0 +1,130 @@
...
+  jsonCrypto: Encrypter with Decrypter
+)(implicit ec: ExecutionContext) {
+
+  def saveForm(saveRequest: SaveRequest): Future[SaveReply] =
+    formRepository.findById(saveRequest.formId).flatMap {
+      case Some(cacheItem) =>
+        val newId = java.util.UUID.randomUUID().toString
+        cacheItem.data.fields.toList
+          .traverse { case (k, v) =>
+            val value = if (k == "form") { v.as[JsObject] - "_id" - "userId" }
```
DursunGundogan, line 41: Can we use `===` instead of `==`?

josef-vlach (approved):

PR #1079, opened 2024-01-16
GFORMS-2512 - Streamline transfer from upscan to object store

DursunGundogan:


app/uk/gov/hmrc/gform/upscan/UpscanController.scala
```diff
@@ -140,22 +142,25 @@ class UpscanController(
...
                          )
                        } else {
-                         fileUploadFrontendAlgebra
-                           .uploadFile(envelopeId, fileId, uploadDetails, compressedFile)
+                         for {
+                           file <- upscanService.download(upscanCallbackSuccess.downloadUrl)
+                           compressedFile =
+                             ImageCompressor.compressIfSupported(file, upscanCallbackSuccess, compression)
+                           res <- fileUploadFrontendAlgebra
+                                    .uploadFile(envelopeId, fileId, uploadDetails, compressedFile)
```
josef-vlach, line 161: Doing the compression is sort of reverting the advantage of using `uploadFromUrl` so we should drop doing it.

DursunGundogan: I would like to keep this compression if the form does not use the object store.

DursunGundogan: Removed

josef-vlach (approved):

PR #1092, opened 2024-02-05
GFORMS-2487: When creating a snapshot the form template should also b…

lxol:
…e added to the snapshot


app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
```diff
@@ -40,17 +51,64 @@ class TestOnlyFormService(
           val newId = java.util.UUID.randomUUID().toString
           cacheItem.data.fields.toList
             .traverse { case (k, v) =>
-              val value = if (k === "form") { v.as[JsObject] - "_id" - "userId" }
-              else v
-              snapshotRepository.put(newId)(DataKey(k), value)
+              if (k === "form") {
+                val formTemplateId = v.as[JsObject].value("formTemplateId").as[String]
+                val prefix = UniqueStringGenerator.generateUniqueString()
```
josef-vlach, line 56: Why not to use `newId` ?

app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
```diff
@@ -40,17 +51,64 @@ class TestOnlyFormService(
...
+              )
+            )
             .map(_ => newId)
         case None =>
           Future.failed(new Exception(s"We could not find cache item with id: ${saveRequest.formId}"))
       }
       .flatMap(snapShotId => getSnapshotData(snapShotId))
 
+  def restoreSnapshotTemplate(snapshotId: String): Future[Unit] = {
+    val snapshotItem = snapshotRepository.findById(snapshotId)
```
josef-vlach, line 95: Why this returns some `CacheItem` which is not our type?

Can `snapshotRepository` operate on our type instead?

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
```diff
@@ -26,7 +26,10 @@ case class Snapshot(
   templateId: String,
   snapshotId: String,
   savedAt: Instant,
-  description: String
+  description: String,
+  gformVersion: String,
+  gformFrontendVersion: String,
+  hasTemplate: Boolean
```
josef-vlach, line 32: This feels like we are creating tech debt in snapshot logic. Snapshot should always have a type.

I think we should migrate user data instead. (Or remove existing snapshots after user consent)

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
lxol: `newId` is too long to use as a prefix. Alex asked for something shorter. I came up with a way to encode the number of seconds since 1 jan 2024 as a alphanumeric string, so we have a new form id as for example `4xumt_tst123`

app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
lxol: We can, yes. But I thought it is unnecessary. We just manipulate jsons that represent `Form` or `Template` in the snapshot  and I thought to do that the functionality that bare mongo cache repository is adequate. We don't need to wrap/unwrap it to our data.

app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
josef-vlach: So lets use first 8 (or 4 if you really insist) letters from `newId`.

lxol: Potentially we might have a conflicts there. My method guarantees the uniqueness as long as it is not generated on the same second.

lxol: and it is max 5-6 chars in the foreseeable future.

app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
josef-vlach: There is at least 5 fields I can see at the moment. Let's use proper type for it.

lxol: ok

app/uk/gov/hmrc/gform/testonly/UniqueStringGenerator.scala
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.gform.testonly
+
+import java.time.Instant
+
+object UniqueStringGenerator {
```
josef-vlach, line 21: We don't need such a complexity.

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
```diff
@@ -16,78 +16,166 @@
...
+
+  def apply(
+    originalForm: Form,
+    originalTemplate: FormTemplateRaw,
+    description: Description,
+    gformVersion: GformVersion,
+    gformFrontendVersion: GformFrontendVersion
+  ): SnapshotItem = {
+    // add prefix to the template id for both saved form and template
+    val prefix = UniqueStringGenerator.generateUniqueString()
```
josef-vlach, line 161: Use `snapshotId` bit instead.

That will have an added benefit that we will be able to identified what snapshot was current session created from (it can come handy)

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
```diff
@@ -16,78 +16,166 @@
...
 import play.api.libs.json._
+import uk.gov.hmrc.crypto.{ Decrypter, Encrypter }
+import uk.gov.hmrc.gform.save4later.EncryptedFormFormat
+import uk.gov.hmrc.gform.sharedmodel.form.Form
+import uk.gov.hmrc.gform.sharedmodel.form.FormData
+import uk.gov.hmrc.gform.sharedmodel.form.FormId
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.FormTemplateId
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.FormTemplateRaw
 
 case class Snapshot(
```
josef-vlach, line 31: Why we need this type? Why not use `SnapshotItem` instead?

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
```diff
@@ -16,78 +16,166 @@
...
   implicit val writes: OWrites[Snapshot] = derived.owrites()
-
 }
 
 case class SnapshotWithData(
   snapshot: Snapshot,
-  formData: JsObject
+  formData: FormData
 )
 object SnapshotWithData {
```
josef-vlach, line 59: Why we need this type? Why not use `SnapshotItem` instead?

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
```diff
@@ -16,78 +16,166 @@
...
+  implicit val format: Format[GformVersion] = Json.valueFormat
+}
+
+case class GformFrontendVersion(value: String) extends AnyVal
+
+object GformFrontendVersion {
+  implicit val format: Format[GformFrontendVersion] = Json.valueFormat
+}
+
+case class SnapshotItem(
```
josef-vlach, line 124: It would be nice to call this simply `Snapshot`

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
lxol: `SnapshotItem` also contains a template. I just wanted a ADT that is not bloated with form data and template but can be usefull to  display  a List[Snapshot]  on the restore page. It will be huge if I have to send List[SnapshotItem]

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
lxol: yes, `SnapshotItem` should work here I think. i will refactor

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
lxol: but how would I call current `Snapshot` that is used for List[Snapshot]

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
lxol: but again the SnapshotItem has template and the whole form, but for this I need only some meta information about snapshot, which is `Snapshot` and formdata

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
lxol: Shall I rename `Snapshot` to `SnapshotMeta` and `SnapshotItem` to `Snapshot`?

josef-vlach: We can call `SnapshotOverview` the class to be used in a List.

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
josef-vlach: Let's call this `SnapshotOverview`

app/uk/gov/hmrc/gform/testonly/SnapshotADTs.scala
```diff
@@ -16,78 +16,158 @@
...
+      description,
+      gformVersion,
+      gformFrontendVersion,
+      snapshotTemplateId,
+      Instant.now
+    )
+  }
+
+  def format(jsonCrypto: Encrypter with Decrypter): Format[Snapshot] = {
+    implicit val formatFormEncrypted: Format[Form] = EncryptedFormFormat.formatEncrypted(jsonCrypto)
```
josef-vlach, line 169: Do we really need to encrypt form data in snapshot?

josef-vlach (changes requested):

josef-vlach (approved):

PR #1094, opened 2024-02-07
GFORMS-2533: Add Conditional Validation to Schema

Aashvin:


conf/formTemplateSchema.json
```diff
@@ -729,7 +729,215 @@
             "$ref": "#/$defs/fields"
           }
         }
-      }
+      },
+      "allOf": [
+        {
+          "anyOf": [
+            {
+              "not": {"required": ["infoText"]}
```
josef-vlach, line 737: Would it be possible to format this? (applies to whole json)

My formatter do this for example:
```
              "not": {
                "required": [
                  "infoText"
                ]
              }
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/formtemplate/JsonSchemeValidator.scala
```diff
@@ -32,20 +34,67 @@ object JsonSchemeValidator {
...
+                }.toList
+
+              errors.map(_.getMessage).toList.slice(0, indexOfAnyOf) ++ formattedConditionalValidationErrors
+            } else {
+              errors.map(_.getMessage).toList
+            }
+          }
+
+          SchemaValidationException(
+            NonEmptyList(conditionalValidationErrorMessages.head, conditionalValidationErrorMessages.tail)
```
lxol, line 95: is that possible that `conditionalValidationErrorMessages is empty?
Also you can use NonEmptyList.fromList method instead here or its unsafe version.

lxol: also `errors` on line 65 is NonEmptyList and the final result is NonEmptyList. May be we should keep it as `Nel` and never convert it to list, to ensure non-emptiness?

lxol: oops, just saw @josef-vlach comments about the same...

test/uk/gov/hmrc/gform/formtemplate/JsonSchemeValidatorSpec.scala
```diff
@@ -0,0 +1,719 @@
...
+  test(
+    "validateJson rejects the form gracefully when the property infoType is used and the type property is not [info]"
+  ) {
+    val jsonTemplate = constructTestOneSectionJsonTemplate(
+      propertyDependencies = Map("type" -> "text"),
+      strings = Map("infoType" -> "noformat"),
+      ints = Map(),
+      arraysOfStrings = Map(),
+      arraysOfObjects = Map(),
+      objects = Map()
```
josef-vlach, line 100: Can we replace all these parameters with actual json under test? Like:
```
json"""
  {
    "infoType": "noformat",
    "type": "text",
    "id": "testId",
    "label": "test label"
  }"""
```
That would make the tests really nice to read.

josef-vlach (changes requested):

josef-vlach (approved):
Please use Squash and merge 🙏

PR #1096, opened 2024-02-12
GFORMS-2550: Conditional Validation for the Format Property

Aashvin:
Adds the conditional validation checks for the format property.

Also adds test cases for different fields in the same section since should be covered.

### DIFF
--- a/app/uk/gov/hmrc/gform/formtemplate/JsonSchemeValidator.scala
+++ b/app/uk/gov/hmrc/gform/formtemplate/JsonSchemeValidator.scala
@@ -84,9 +84,9 @@ object JsonSchemeValidator {
                   val deduplicatedConditionalRequirements: List[String] = {
                     val distinctProperties = conditionalRequirements(errorProperty).map(_.getRequiredProperty).distinct
                     val groupedProperties = conditionalRequirements(errorProperty).groupBy(_.getRequiredProperty)
-                    distinctProperties.map(property => property -> groupedProperties(property)).map {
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/formtemplate/JsonSchemeValidator.scala
```diff
@@ -78,13 +77,21 @@ object JsonSchemeValidator {
...
 
               val formattedConditionalValidationErrors: List[String] =
                 deduplicatedConditionalValidationErrors.map { case (errorProperty, errorLocation) =>
-                  s"$errorLocation: Property $errorProperty can only be used with ${conditionalRequirements(errorProperty)
-                    .mkString(", ")}"
-                }.toList
+                  val deduplicatedConditionalRequirements: List[String] = {
+                    val distinctProperties = conditionalRequirements(errorProperty).map(_.getRequiredProperty).distinct
+                    val groupedProperties = conditionalRequirements(errorProperty).groupBy(_.getRequiredProperty)
+                    distinctProperties.map(propertyKey => propertyKey -> groupedProperties(propertyKey)).map {
```
Aashvin, line 87: `val distinctProperties` is used to preserve the order of required properties found in `private val conditionalRequirements`, since `groupBy` does not preserve order of keys - makes the determinism easier to predict. Happy to remove this if necessary and just use `groupBy`.

josef-vlach (approved):

PR #1101, opened 2024-02-20
GFORMS-2552: Add JSON Schema Check to Prevent a Prefix Being Added to a Field with Sterling Format

Aashvin:


conf/formTemplateSchema.json
```diff
@@ -965,6 +965,19 @@
...
           }
+        },
+        "prefix": {
+          "required": [
+            "type"
+          ],
+          "properties": {
+            "type": {
+              "pattern": "^text$"
+            },
```
Aashvin, line 976: I've added the requirement for the type field since this is the case, although it is not on the ticket. Happy to remove if necessary.

josef-vlach (approved):

PR #1104, opened 2024-02-21
GFORMS-2518: Add a delete link to remove a snapshot that is no longer…

lxol:
… required

### DIFF
--- a/app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
+++ b/app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
@@ -335,7 +335,7 @@ class TestOnlyController(
   def deleteSnapshot(snapshotId: SnapshotId) = Action.async { _ =>
     testOnlyFormService
       .deleteSnapshot(snapshotId)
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/gform/testonly/TestOnlyController.scala
```diff
@@ -331,6 +331,12 @@ class TestOnlyController(
       val updateRequest: UpdateFormDataRequest = request.body
       testOnlyFormService.updateFormData(updateRequest).map(saveReply => Ok(Json.toJson(saveReply)))
     }
+
+  def deleteSnapshot(snapshotId: SnapshotId) = Action.async { _ =>
+    testOnlyFormService
+      .deleteSnapshot(snapshotId)
+      .map(_ => Ok(""))
```
josef-vlach, line 338: Use `NoContent` if you don't want send any response.

josef-vlach (changes requested):

josef-vlach (approved):

PR #1131, opened 2024-03-25
GFORMS-2483: Add a link to view form data in toolbox link

lxol:


app/uk/gov/hmrc/gform/testonly/TestOnlyFormService.scala
```diff
@@ -70,6 +70,11 @@ class TestOnlyFormService(
     FormTemplateRaw(updatedValue)
   }
 
+  def getFormData(formId: FormId)(implicit hc: HeaderCarrier): Future[FormData] =
+    formMongoCache
+      .get(formId)
+      .map(_.formData)
```
josef-vlach, line 76: Let's print whole form.

josef-vlach (changes requested):
I don't think we need any new endpoint. I think printing `Json.toJson(cache.form)` in frontend should be enough.

PR #1135, opened 2024-04-04
GFORMS-2610: Allow for different complete sentences to be defined for…

lxol:
… smart string properties to facilitate Welsh translation


app/uk/gov/hmrc/gform/sharedmodel/SmartString.scala
```diff
@@ -122,11 +126,120 @@ object SmartStringTemplateReader {
       ss              <- buildSmartStringFromLocalisedStringAndInterpolations(localisedString, interpolations)
     } yield ss
 
+  private def readFromTemplateArray(array: JsArray): JsResult[SmartString] = {
+    val jsObjects = array.value.collect { case jsObj: JsObject => jsObj }
+    val abc = jsObjects.map { jsObj =>
```
josef-vlach, line 131: better name

josef-vlach (changes requested):
Interesting solution. I'm not sure if this is the right way around this.

Two critical observations:
1. `$n` doesn't work in ATL
2. Markdown rendering is broken

app/uk/gov/hmrc/gform/sharedmodel/SmartString.scala
lxol: I just wanted it to be in line with other `readFromTemplate...` methods here. It is play-json String, Object or Array
```  val templateReads: Reads[SmartString] = new Reads[SmartString] {
    override def reads(json: JsValue): JsResult[SmartString] = json match {
      case JsString(value) => readFromTemplateString(value)
      case obj: JsObject   => readFromTemplateObject(obj.fields.toMap)
      case arr: JsArray =>
        readFromTemplateArray(arr)
      case _ => JsError(s"Expected a String or an Object while reading a SmartString. Got $json")
    }
```

lxol:
> Interesting solution. I'm not sure if this is the right way around this.
> 
> Two critical observations:
> 
> 1. `$n` doesn't work in ATL
> 2. Markdown rendering is broken

Should be okay with a new fix....
I considered making SmartString  a coproduct with SmartString and SmartStringIf  extending it, but it appeared to me too intrusive.

app/uk/gov/hmrc/gform/sharedmodel/SmartString.scala
josef-vlach: I was referring to `abc`

lxol: changed it....

josef-vlach:
Have you tried to redefine `SmartString` as a simple coproduct?

```
sealed trait SmartString

object SmartString {
  case class Single(localised: LocalisedString, interpolations: List[Expr]) extends SmartString
  case class Multiple(smarts: NonEmptyList[(LocalisedString, List[Expr], IncludeIf)]) extends SmartString
}
```

This looks more idiomatic to me than extending `Expr`.
Wdyt?

lxol:
> Have you tried to redefine `SmartString` as a simple coproduct?

Yes, I played around with something like this. It is definitely doable, but I thought that considering smart string with if as if it is a normal smart string but with different representation of the interpolations ( which is essentially what it  is) would be easier. 
Let's discuss tomorrow.

PR #1136, opened 2024-04-04
GFORMS-2611: Allow a caption to be defined for a task and use this for all sections unless a caption is defined on section within task

Aashvin:


app/uk/gov/hmrc/gform/formtemplate/Rewriter.scala
```diff
@@ -533,18 +534,71 @@ trait Rewriter {
...
+                  redirects = replaceRedirects(page.redirects),
+                  confirmation = replaceConfirmation(page.confirmation)
+                )
+              ),
+              fields = replaceFieldsNel(s.fields),
+              cyaPage = s.cyaPage.map(replaceCheckYourAnswersPage),
+              caption = newCaption
+            )
+        }
+
```
josef-vlach, line 589: Can we update only a caption and then call `updateClassicSection` to reduce duplication?

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/builder/BuilderController.scala
```diff
@@ -213,14 +213,23 @@ object BuilderSupport {
...
 
+//  def modifyTaskCaption(json: Json, taskSectionData: Json, sectionPath: SectionPath): Json = {
+//    val propertyList = List(
+//      Property("caption", PropertyBehaviour.PurgeWhenEmpty)
+//    )
+//
+//    val history = sectionPath.asHistory
+//    updateJsonByPropertyList(propertyList, json, taskSectionData, history)
+//  }
+
```
josef-vlach, line 232: Please delete this if it is not needed

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/builder/BuilderController.scala
Aashvin: Apologies, deleted.

josef-vlach (approved):

PR #1141, opened 2024-04-09
GFORMS-2619: Allow a Caption to be Defined for ATL and use this for all Pages Unless a Caption is Defined on the Page in ATL

Aashvin:


app/uk/gov/hmrc/gform/formtemplate/Rewriter.scala
```diff
@@ -537,23 +540,10 @@ trait Rewriter {
       def updateTaskSectionCaptions(section: Section, taskCaption: Option[SmartString]): Section =
         section match {
           case s: Section.NonRepeatingPage =>
-            val newCaption: Option[SmartString] = s.page.caption match {
-              case Some(value) => Some(value)
-              case None        => taskCaption
-            }
-            s.copy(page = s.page.copy(caption = newCaption))
+            s.copy(page = s.page.copy(caption = if (s.page.caption.nonEmpty) s.page.caption else taskCaption))
```
josef-vlach, line 543: I think this can be more concise version:
```suggestion
            s.copy(page = s.page.copy(caption = s.page.caption.orElse(taskCaption)))
```

The same holds for other usages of `.nonEmpty`

josef-vlach (approved):

PR #1156, opened 2024-05-03
GFORMS-2678 - Prod deploymeny fails when we restart the gform

DursunGundogan:


app/uk/gov/hmrc/gform/formstatistics/FormStatisticsModule.scala
```diff
@@ -40,10 +38,7 @@ class FormStatisticsModule(
     new Repo[Form](
       "forms",
       mongoModule.mongoComponent,
-      _._id.value,
```
lxol, line 43: I think you will have the same error. You create a new PlayMongoReposiory with empty indexes and it runs default ensureIndexes on construction. `ensureIndexes` then checks that existing indexes in `gform.forms` are not equal to empty indexes and throws the same error.

lxol: Not sure what to do here. May be we need to just duplicate the same indexes in the Statistics repo. Or somehow have it set up in one place.

lxol (approved):

PR #1158, opened 2024-05-13
GFORMS-2263: Enable Radio and Checkbox Labels to be Updated

Aashvin:


app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala
```diff
@@ -580,7 +580,8 @@ object FormTemplateValidator {
   }
 
   def validateChoiceNoneChoiceAndError(sectionsList: List[Page]): ValidationResult = {
-    def check(choice: Choice): Boolean = choice.noneChoice.isDefined != choice.noneChoiceError.isDefined
+    def check(choice: Choice): Boolean =
+      choice.`type` == Checkbox && choice.noneChoice.isDefined != choice.noneChoiceError.isDefined
```
DursunGundogan, line 584: Can we use the safe equalities like === and =!= instead of == and !=?

josef-vlach (approved):

PR #1161, opened 2024-05-15
GFORMS-2671: The substring Function Should be Supported in booleanExpressions and Support user.enrolments

Aashvin:


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
Aashvin, line 1: I would think commas aren't allowed in the service or identifier names but please do correct me if I'm wrong.

josef-vlach (approved):

PR #1170, opened 2024-06-11
GFORMS-2686: Include standard elements required by robotics in xml file produced with handlebar template

Aashvin:


app/uk/gov/hmrc/gform/submission/PdfAndXmlSummariesFactory.scala
```diff
@@ -172,10 +173,46 @@ object PdfAndXmlSummariesFactory {
...
+
+      val indentedHandlebar = wrapAndIndentHandlebarXml(None, formattedHandlebar)
+
+      XmlGeneratorService.xmlDec + "\n" + wrapAndIndentHandlebarXml(
+        Some("<data xmlns:xfa=\"http://www.xfa.org/schema/xfa-data/1.0/\">"),
+        indentedHandlebar
+      )
+    }
+
+    private def wrapAndIndentHandlebarXml(wrappingTag: Option[String], contents: String): String =
```
lxol, line 209: would it be better to use `scala.xml.PrettyPrinter`  for xml formatting?

Aashvin: I was looking into this but apparently it's not thread safe. If that doesn't matter then I'll be happy to make the change to use it

Aashvin: After discussion, will change to use `PrettyPrinter`

lxol (approved):

PR #1207, opened 2024-09-10
GFORMS-2861 - Do not allow empty spaces is choices value property. Re…

jmartinhmrc:
…turned readable error message for the pattern's regex.

### DIFF
--- a/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
@@ -397,7 +397,7 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
     str
   }
 
-  lazy val anything: Parser[String] = """.*""".r ^^ { str => str }
+  lazy val notEmpty: Parser[String] = """.+""".r ^^ { str => str }
 
   lazy val alphaNumericWithSpace: Parser[String] = """[a-zA-Z\d ]*""".r ^^ { str =>
     str
@@ -620,7 +620,7 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
       OptionDataValue.FormCtxBased(formCtx)
     } | alphabeticOnly ~ expr ^^ { case prefix ~ expr =>
       OptionDataValue.ExprBased(prefix, expr)
-    } | anything ^^ (str => OptionDataValue.StringBased(str))
+    } | notEmpty ^^ (str => OptionDataValue.StringBased(str))
 }
 
 object ValueParser extends ValueParser with PackratParsingHelper {
--- a/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
@@ -385,7 +385,7 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
     str
   }
 
-  lazy val alphabeticWithSpacesAndCommas: Parser[String] = """[a-zA-Z\s,]*""".r ^^ { str =>
+  lazy val alphaNumericWithSpacesCommasAndUnderscores: Parser[String] = """[a-zA-Z\s\d_,]*""".r ^^ { str =>
     str
   }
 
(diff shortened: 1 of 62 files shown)


conf/formTemplateSchema.json
```diff
@@ -758,7 +758,8 @@
                     "type": "string"
                   },
                   "value": {
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^\\S+$"
```
josef-vlach, line 762: We cannot reject whitespaces altogether like this. Problem is that `value` can also include arbitrary expressions and so we need to do the whitespace check on more granular level. See parsing of `value` for more [clarity](https://github.com/hmrc/gform/blob/main/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala#L604-L610).

I'd say this granular check should be implemented as another check in [Verifier.scala](https://github.com/hmrc/gform/blob/main/app/uk/gov/hmrc/gform/formtemplate/Verifier.scala)

josef-vlach (changes requested):

conf/formTemplateSchema.json
jmartinhmrc: Thanks for this info! Implementation has been changed

app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -606,7 +610,7 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
       OptionDataValue.FormCtxBased(formCtx)
     } | alphabeticOnly ~ expr ^^ { case prefix ~ expr =>
       OptionDataValue.ExprBased(prefix, expr)
-    } | formatParserAlphabeticOnly ^^ (str => OptionDataValue.StringBased(str))
+    } | alphabeticWithSpacesAndCommas ^^ (str => OptionDataValue.StringBased(str))
```
josef-vlach, line 613: This change may cause forms to get broken. For example valid value like `op2` will be broken by this. Can we just not this change?

josef-vlach (dismissed):

app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
jmartinhmrc: I can update the regex to allow the conditions of the original word character matcher so it's like this:
`[a-zA-Z\s\d_,]`
Then commas and spaces will be allowed as well and rejected by the FormTemplateValidator. What do you think?

jmartinhmrc: From what I've looked at - If I don't change this regex, when the user adds non-expression values with spaces, the OptionData for the choice will be initialised as index based and will have no value property to check in FormTemplateValidator. So the user remains unaware of the issue. 
Do you have any thoughts on another way around this? As the current error (if you leave one value with a space by accident or otherwise) is the same as you mentioned:
`Choice component has some options with 'value' and some options without 'value'`

jmartinhmrc: Maybe throw exception like is done [here](https://github.com/hmrc/gform/blob/a4a45c0ba0df2a6337720ac908bf78c31311db13/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/OptionDataValue.scala#L32)

app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -397,6 +397,8 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
     str
   }
 
+  lazy val notEmpty: Parser[String] = """.+""".r ^^ { str => str }
```
DursunGundogan, line 400: Do we really need to add this regex? I think we should accept only Alphabetic chars.

DursunGundogan: We should think how to handle empty values. It returns a parser error. 
```           {
              "en": "Blind Persons Allowance",
              "value": ""
            },```

jmartinhmrc: Ok thanks, I'll look into this!

jmartinhmrc: Hi @DursunGundogan, I've updated it to allow any String based values to the template validator and added an additional check for empty string or spaces.

DursunGundogan (approved):

PR #1208, opened 2024-09-11
GFORMS-2389 - Make all properties that accept true/false or yes/no to…

DursunGundogan:
… be boolean


app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DraftRetrievalMethod.scala
```diff
@@ -17,26 +17,19 @@
 package uk.gov.hmrc.gform.sharedmodel.formtemplate
 
 import play.api.libs.json._
-import uk.gov.hmrc.gform.formtemplate.FormComponentMakerService.{ IsFalseish, IsTrueish }
 
 sealed trait DraftRetrievalMethod
 
 case class OnePerUser(continueOrDeletePage: ContinueOrDeletePage) extends DraftRetrievalMethod
 case class FormAccessCodeForAgents(continueOrDeletePage: ContinueOrDeletePage) extends DraftRetrievalMethod
-case object BySubmissionReference extends DraftRetrievalMethod
```
josef-vlach, line 26: Why is this removed?

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ContinueOrDeletePage.scala
```diff
@@ -28,9 +27,9 @@ object ContinueOrDeletePage {
...
   implicit val format: OFormat[ContinueOrDeletePage] = derived.oformat()
 
-  def unapply(str: String): Option[ContinueOrDeletePage] = str match {
-    case IsFalseish() => Some(Skip)
-    case IsTrueish()  => Some(Show)
-    case _            => None
+  def unapply(str: Boolean): Option[ContinueOrDeletePage] = if (str) {
+    Some(Show)
+  } else {
+    Some(Skip)
```
josef-vlach, line 33: This never returns `None` so there is no need for `Option[ContinueOrDeletePage]` as a return type

josef-vlach (changes requested):

josef-vlach:
Also would it be possible to get rid of these objects?

```
object IsTrueish
object IsFalseish
final object IsDisplayCharCountFalse
final object IsDisplayCharCountTrue
```

app/uk/gov/hmrc/gform/sharedmodel/formtemplate/ContinueOrDeletePage.scala
DursunGundogan: There is a pattern matching [here](https://github.com/hmrc/gform/blob/main/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DraftRetrievalMethod.scala#L31) I think we should return it as Option

josef-vlach: > There is a pattern matching [here](https://github.com/hmrc/gform/blob/main/app/uk/gov/hmrc/gform/sharedmodel/formtemplate/DraftRetrievalMethod.scala#L31) I think we should return it as Option

In provided link we can get rid of matching on `showContinueOrDeletePage: Boolean` and use id directly.

So we should provide this 

```
def fromBoolean(str: Boolean): ContinueOrDeletePage = if (str) Show else Skip
```
instead of this `unapply` method

josef-vlach (approved):

PR #1212, opened 2024-09-13
GFORMS-2885 - Improve error message when posting page/section with "c…

DursunGundogan:
…y": ""

### DIFF
--- a/app/uk/gov/hmrc/gform/core/parsers/FormatParser.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/FormatParser.scala
@@ -198,7 +198,7 @@ trait FormatParser extends ValueParser {
 
   case class WholeFractional(w: Int, f: Int)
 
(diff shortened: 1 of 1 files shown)


josef-vlach (commented):
I'm not sure this is fixing the issue?

Thinking about GFORMS-2885 again and given that this is valid
```
{
  "title": {
    "en": ""
  },
  "fields": [
    {
      "id": "email",
      "label": "Email",
      "type": "text",
      "format": "email"
    }
  ]
}
```
I think that AC1 goal of introducing error `section/page with text component email cannot have an empty cy title` is not consistent with how we treat empty smartstring in general see my example with empty `title`.

I think we should simply ignore empty `"cy": ""` ie. we should remove 
```
|| internalSmartString.rawValue(LangADT.Cy).trim === ""
```

from [here](https://github.com/hmrc/gform/blob/main/app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala#L755)

PR #1240, opened 2024-10-25
GFORMS-2906 - Add the ability to look up a rate from a csv

cmanson:


app/uk/gov/hmrc/gform/sharedmodel/DataRetrieveDefinition.scala
```diff
@@ -337,6 +337,46 @@ object DataRetrieveDefinitions {
...
+    ),
+    List(
+      Parameter("regime"),
+      Parameter("code"),
+      Parameter("date")
+    ),
+    Map(
+      DataRetrieve.Attribute("regime")    -> DataRetrieve.AttrType.String,
+      DataRetrieve.Attribute("code")      -> DataRetrieve.AttrType.String,
+      DataRetrieve.Attribute("rate")      -> DataRetrieve.AttrType.String,
```
josef-vlach, line 374: This can't be `String`, otherwise expression like `dataRetrieve.todaysRate.rate` would evaluate to a `StringResult` and `${dataRetrieve.todaysRate.rate * 1.2}` would not work

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/DataRetrieveDefinition.scala
cmanson: I've now added a BigDecimal type

cmanson: Ah okay - I've just tested this with the changes and it certainly works now.

app/uk/gov/hmrc/gform/sharedmodel/DataRetrieve.scala
```diff
@@ -101,6 +101,7 @@ object DataRetrieve {
     case object String extends AttrType
     case object Integer extends AttrType
     case object Date extends AttrType
+    case object BigDecimal extends AttrType
```
josef-vlach, line 104: From the look of it `BigDecimal` behaves exactly as `Integer` (looking at gform-frontend). 

I think we should rename `Integer` (line 102) to a `Number` instead rather than having two types which behaves exactly the same.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/sharedmodel/DataRetrieve.scala
cmanson: Yeah okay, that makes sense. Will get on it.

cmanson: Hi @josef-vlach, I've renamed this now and pushed the changes up.

josef-vlach (approved):

PR #1251, opened 2024-11-19
GFORMS-2429 - Choice expression should be syntactic sugar for concat …

jmartinhmrc:
…expression internally

### DIFF
--- a/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
+++ b/app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
@@ -329,8 +329,8 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
     | "concat(" ~ _expr1 ~ (("," ~> _expr1) *) ~ ")" ^^ { case _ ~ expr ~ exprs ~ _ =>
       Concat(expr +: exprs)
     }
-    | "choice(" ~ FormComponentId.unanchoredIdValidation ~ ")" ^^ { case _ ~ value ~ _ =>
-      ChoiceLabel(FormComponentId(value))
+    | "choice(" ~ _expr1 ~ (("," ~> _expr1) *) ~ ")" ^^ { case _ ~ expr ~ exprs ~ _ =>
+      ChoiceLabel(expr +: exprs)
     }
     | "choicesSelected(" ~ FormComponentId.unanchoredIdValidation ~ ")" ^^ { case _ ~ value ~ _ =>
       ChoicesSelected(FormComponentId(value))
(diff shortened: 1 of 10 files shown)


app/uk/gov/hmrc/gform/core/parsers/ValueParser.scala
```diff
@@ -329,8 +329,8 @@ trait ValueParser extends RegexParsers with PackratParsers with BasicParsers {
     | "concat(" ~ _expr1 ~ (("," ~> _expr1) *) ~ ")" ^^ { case _ ~ expr ~ exprs ~ _ =>
       Concat(expr +: exprs)
     }
-    | "choice(" ~ FormComponentId.unanchoredIdValidation ~ ")" ^^ { case _ ~ value ~ _ =>
-      ChoiceLabel(FormComponentId(value))
+    | "choice(" ~ _expr1 ~ (("," ~> _expr1) *) ~ ")" ^^ { case _ ~ expr ~ exprs ~ _ =>
+      ChoiceLabel(expr +: exprs)
```
josef-vlach, line 333:
```suggestion
    | "choice(" ~ FormComponentId.unanchoredIdValidation ~ ")" ^^ { case _ ~ value ~ _ =>
      Concat(FormCtx(FormComponentId(value)) :: Nil)
```

We should not change the parser.

We should remove `final case class ChoiceLabel(formComponentId: FormComponentId) extends Expr` altogether.

josef-vlach (changes requested):

test/uk/gov/hmrc/gform/core/parsers/ValueParserSpec.scala
```diff
@@ -29,7 +29,7 @@ import uk.gov.hmrc.gform.sharedmodel.{ AvailableLanguages, DataRetrieve, DataRet
 import uk.gov.hmrc.gform.sharedmodel.email.LocalisedEmailTemplateId
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.InternalLink.{ NewForm, NewFormForTemplate, NewSession, PageLink, SignOut }
 import uk.gov.hmrc.gform.sharedmodel.formtemplate.UserField.Enrolment
-import uk.gov.hmrc.gform.sharedmodel.formtemplate._
+import uk.gov.hmrc.gform.sharedmodel.formtemplate.{ TextExpression, _ }
```
josef-vlach, line 32: This is unnecessary change.

josef-vlach (changes requested):

test/uk/gov/hmrc/gform/core/parsers/ValueParserSpec.scala
jmartinhmrc: Ah yeah - thanks

josef-vlach (approved):

PR #1256, opened 2024-11-28
GFORMS-2593 - Support choices and format for addAnotherQuestion property in ATL

cmanson:


app/uk/gov/hmrc/gform/formtemplate/FormTemplateValidator.scala
```diff
@@ -1336,6 +1336,33 @@ object FormTemplateValidator {
...
+
+    def checkChoiceOptions(addToList: Section.AddToList): List[ValidationResult] = {
+      implicit val l: LangADT = LangADT.En
+      val invalid: Invalid = Invalid(
+        s"AddToList '${addToList.title.defaultRawValue}' addAnotherQuestion must only contain choices of 'Yes' and 'No' in that order."
+      )
+
+      def isValid(t: Boolean): ValidationResult = if (t) Valid else invalid
+
+      val choice: Choice = addToList.addAnotherQuestion.`type`.cast[Choice].get
```
DursunGundogan, line 1349: You should avoid using .get, it's an unsafe operator. Can we safely attempt to cast and access options like 
`.`type`.cast[Choice] match {
      case Some(choice) =>`

cmanson: Hi Dursun, yes you're right. I only used .get because it was used elsewhere in the code for the same context - because addAnotherQuestion is always a Choice. I can change it, but I guess it should be changed everywhere.

cmanson: Actually, it's only used elsewhere in the frontend. So I'm going to put additional validation in to ensure that addAnotherQuestion is a Choice component. You can actually submit it with other component types at the moment... and the frontend crashes when you get to that page!

DursunGundogan (approved):

PR #1257, opened 2024-12-02
GFORMS-3051 - Remove Nino enrolments from datastore user session json…

jmartinhmrc:
… for CT submissions


app/uk/gov/hmrc/gform/submission/DataStoreFileGenerator.scala
```diff
@@ -63,4 +63,13 @@ object DataStoreFileGenerator {
 
     (dataStoreJson ++ gformJson.as[JsObject]).toString()
   }
+
+  def filterEnrolments(regime: String, enrolments: List[JsValue]): List[JsValue] =
+    if (regime == "CT") {
```
DursunGundogan, line 68: Can we use === operator?

jmartinhmrc: Updated!

PR #1284, opened 2025-01-17
GFORMS-2733 - Download previously submitted PDF

cmanson:


app/uk/gov/hmrc/gform/sharedmodel/form/Form.scala
```diff
@@ -39,7 +39,8 @@ case class Form(
   thirdPartyData: ThirdPartyData,
   envelopeExpiryDate: Option[EnvelopeExpiryDate],
   componentIdToFileId: FormComponentIdToFileIdMapping,
-  taskIdTaskStatus: TaskIdTaskStatusMapping
+  taskIdTaskStatus: TaskIdTaskStatusMapping,
+  submitted: Option[SubmittedDate]
```
josef-vlach, line 43: Why is this field needed? Why is `Submission.submittedDate` not enough?

cmanson: I did it like this so I wouldn't have to hit 2 Mongo tables, when we're retrieving the form data anyway. Otherwise we'd need to double the number of queries to the database. One to 'forms' collection and one to 'submission' collection. My background is in financial services and performance has always been one of the highest priorities, so I always think of that. Happy to change it if you think we shouldn't do it this way.

josef-vlach: We should have only one source of truth. On frontend I can see this ticked uses sometimes this new field and sometimes `Submission.submittedDate` and that is confusing. So let's drop this field. 

On the frontend we need that second call only for forms which are already submitted, so performance hit for second api call/db hit is going to be minimal.

cmanson: Okay I've removed that additional attribute from the form and am now using submission only.

josef-vlach (approved):

PR #1308, opened 2025-03-10
GFORMS-3011 - Integrate with Information Archive via EIS CMA file tra…

DursunGundogan:


app/uk/gov/hmrc/gform/submission/destinations/DmsSubmitter.scala
```diff
@@ -86,26 +84,7 @@ class DmsSubmitter(
...
-    */
-  private def logFileSizeBreach(envelopeId: EnvelopeId, files: List[File]) = {
-    val totalFileSize = files.map(_.length).sum
-    val totalFileSizeMB = Math.ceil(totalFileSize / 1000000.toDouble)
-    val thresholdMB = 100
-    logger.info(
-      s"GForm DMS submission attachments size is $totalFileSize B (rounded to $totalFileSizeMB MB) [envelopeId=${envelopeId.value}]"
-    )
-    if (totalFileSizeMB > thresholdMB) {
-      logger.warn(s"GForm DMS submission attachments size exceeds $thresholdMB MB [envelopeId=${envelopeId.value}]")
```
josef-vlach, line 108: This is used [here](https://github.com/hmrc/alert-config/blob/main/src/main/scala/uk/gov/hmrc/alertconfig/configs/Gforms.scala#L21). 

Why are are we removing it?

DursunGundogan: After object-store integration, I thought this check was unnecessary. I think there is filesize control in frontend. Also  threshold value is hardcoded. val thresholdMB = 100

DursunGundogan: Reverted

josef-vlach (approved):

PR #1323, opened 2025-03-25
GFORMS-3190 - Add a feature to allow a total to be displayed on ATL repeater page when Key Value pairings are used

jmartinhmrc:


conf/formTemplateSchema.json
```diff
@@ -469,6 +469,13 @@
             }
           ]
         },
+        "descriptionTotal": {
+          "anyOf": [
+            {
+              "$ref": "#/$defs/atlDescriptionObject"
+            }
+          ]
```
josef-vlach, line 477:
```suggestion
          "$ref": "#/$defs/atlDescriptionObject"
```

Can't this be just single `$ref`

app/uk/gov/hmrc/gform/translation/TextExtractor.scala
```diff
@@ -338,6 +338,7 @@ object TextExtractor {
       Pure(DownField("progressIndicator")) :: root,
       Pure(DownField("summaryName")) :: root,
       Pure(DownField("description")) :: root,
+      Pure(DownField("descriptionTotal")) :: root,
```
josef-vlach, line 341: This is not correct.

`descriptionTotal` doesn't contain any text for translation.

It needs to be (pseudocode follows): 
- `descriptionTotal \ key`
- `descriptionTotal \ value`

josef-vlach: Note. Yes `description` is wrong also (in the case it is used as `key/value`).

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/translation/TextExtractor.scala
jmartinhmrc: Thanks - updated

josef-vlach (approved):

PR #1327, opened 2025-04-03
GFORMS-3010 - Facilitate additional dynamic headers for profile

cmanson:


conf/application.conf
```diff
@@ -257,6 +257,12 @@ microservice {
             host = localhost
             port = 9197
             base-path = "/gform-stubs"
+            http-headers {
+                "X-Transmitting-System" = "CMA"
+                "X-Originating-System" = "MDTP"
+                "correlationid" = "{UUID}"
```
josef-vlach, line 263: Shouldn't `correlationid` header be an our `envelopeId`?

What is the value of generating random UUID we won't be able to know where it comes from.

josef-vlach: This will be handle in new ticket

josef-vlach (approved):

PR #1333, opened 2025-04-08
GFORMS-3010 - Integrate with EIS API for creation of ETMP charge

cmanson:


josef-vlach (commented):
I find these changes to introduce way too much complexity then is needed. 

I think if we changed

```
private val httpClientMap: Map[ProfileName, HttpClient[FOpt]] =
```

to

```
private val httpClientMap: Map[ProfileName, EnvelopeId => HttpClient[FOpt]] =
```

this would lead to cleaner implementation.

Have you considered such an approach?

cmanson:
> I find these changes to introduce way too much complexity then is needed.
> 
> I think if we changed
> 
> ```
> private val httpClientMap: Map[ProfileName, HttpClient[FOpt]] =
> ```
> 
> to
> 
> ```
> private val httpClientMap: Map[ProfileName, EnvelopeId => HttpClient[FOpt]] =
> ```
> 
> this would lead to cleaner implementation.
> 
> Have you considered such an approach?

Hi Josef - no I hadn't. My scala skills aren't quite at that level yet - my brain still works quite procedurally. I'll look into it now.

app/uk/gov/hmrc/gform/wshttp/HttpClient.scala
```diff
@@ -56,30 +56,31 @@ object HttpClient {
...
-        httpRequest(underlying.post(_, _)(addContentTypeHeader(hc, contentType)), uri, xmlString)(None)
+        def post(uri: String, xmlString: String)(implicit hc: HeaderCarrier): F[HttpResponse] =
+          httpRequest(underlying.post(_, _)(addContentTypeHeader(hc, contentType)), uri, xmlString)(None)
 
-      def put(uri: String, xmlString: String)(implicit hc: HeaderCarrier): F[HttpResponse] =
-        httpRequest(underlying.put(_, _)(addContentTypeHeader(hc, contentType)), uri, xmlString)(None)
-    }
+        def put(uri: String, xmlString: String)(implicit hc: HeaderCarrier): F[HttpResponse] =
+          httpRequest(underlying.put(_, _)(addContentTypeHeader(hc, contentType)), uri, xmlString)(None)
+      }
```
josef-vlach, line 83: Would it be possible to undo these changes?

josef-vlach (changes requested):

josef-vlach (approved):

DursunGundogan (approved):

PR #1334, opened 2025-04-08
GFORMS-3322 - ATL property descriptionTotal not allowing top level expressions in smart strings

jmartinhmrc:


app/uk/gov/hmrc/gform/formtemplate/Substituter.scala
```diff
@@ -419,6 +419,16 @@ object Substituter {
...
+  implicit def descriptionTotalSubstituter[A](implicit
+    ev: Substituter[A, Expr],
+    ev2: Substituter[A, BooleanExpr]
+  ): Substituter[A, Option[AtlDescription.KeyValueBased]] = (substitutions, t) =>
+    t match {
+      case Some(kvPair) =>
+        Option(kvPair.copy(key = kvPair.key(substitutions), value = kvPair.value(substitutions)))
+      case _ => None
+    }
+
```
josef-vlach, line 431:
```suggestion
  implicit def descriptionTotalSubstituter[A](implicit
    ev: Substituter[A, Expr],
    ev2: Substituter[A, BooleanExpr]
  ): Substituter[A, AtlDescription.KeyValueBased] = (substitutions, t) =>
    t.copy(key = t.key(substitutions), value = t.value(substitutions))
```

There is dedicated implicit for handling `Option` so you can just write ☝️

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/formtemplate/Substituter.scala
jmartinhmrc: thanks @josef-vlach!  Updated

josef-vlach (approved):

PR #1343, opened 2025-04-23
GFORMS-3331 - Only allow minimumCL to be 200 or 250 for hmrcVerfied a…

jmartinhmrc:
…uth config


DursunGundogan:
Can we manage it using the form template schema?

DursunGundogan (approved):

PR #1489, opened 2026-04-14
GFORMS-4103: NRS work items metadata support to enable search

PavelVjalicin:
The data is stored like this now:

```
"item": {
    "envelopeId": "5f9a254f-a3cd-47b2-ba05-ecbb72c38e7e",
    "formTemplateId": "nrs-orchestrator-test",
    "submissionRef": "3ZLE-1RYN-LTYA",
    "destinationId": "nrsOrchestrator"
    "data": "Q1b9KNFmn7Go61j0h+Y/kzcYEoF..."
  }
```
  
Added a temp fallback reads method that fills in empty values as empty strings for old data structure.


app/uk/gov/hmrc/gform/scheduler/nrsOrchestrator/NrsOrchestratorWorkItem.scala
```diff
@@ -36,8 +38,63 @@ final case class NrsOrchestratorWorkItem(
...
+  private val logger = LoggerFactory.getLogger(getClass)
+  implicit val envelopeIdFormat: Format[EnvelopeId] = EnvelopeId.vformat
+  implicit val formTemplateIdFormat: Format[FormTemplateId] = FormTemplateId.vformat
+  implicit val submissionRefFormat: Format[SubmissionRef] = SubmissionRef.vformat
+  def formatEncrypted(implicit jsonCrypto: Encrypter with Decrypter): Format[NrsOrchestratorWorkItem] = {
+    implicit val encryptedDataFormat: Format[NrsOrchestratorWorkItemData] =
+      EncryptedFormat.formatEncrypted(jsonCrypto)(Json.format[NrsOrchestratorWorkItemData])
+    new Format[NrsOrchestratorWorkItem] {
+      override def reads(json: JsValue): JsResult[NrsOrchestratorWorkItem] =
+        Json.reads[NrsOrchestratorWorkItem].reads(json) match {
```
josef-vlach: This is macro call. We should assign it to a `val` and reuse it, rather than evaluating the macro on every call.

```
Json.reads[NrsOrchestratorWorkItem]
```

We can create `Format` and reuse it in both `reads` and `writes`:
```
val format: Format[NrsOrchestratorWorkItem] = Json.format[NrsOrchestratorWorkItem]
```

app/uk/gov/hmrc/gform/scheduler/nrsOrchestrator/NrsOrchestratorAttachmentWorkItem.scala
```diff
@@ -16,20 +16,72 @@
...
+  private val logger = LoggerFactory.getLogger(getClass)
+  implicit val envelopeIdFormat: Format[EnvelopeId] = EnvelopeId.vformat
+  implicit val formTemplateIdFormat: Format[FormTemplateId] = FormTemplateId.vformat
+  implicit val submissionRefFormat: Format[SubmissionRef] = SubmissionRef.vformat
+  def formatEncrypted(implicit jsonCrypto: Encrypter with Decrypter): Format[NrsOrchestratorAttachmentWorkItem] = {
+    implicit val encryptedDataFormat: Format[NrsOrchestratorAttachmentWorkItemData] =
+      EncryptedFormat.formatEncrypted(jsonCrypto)(Json.format[NrsOrchestratorAttachmentWorkItemData])
+    new Format[NrsOrchestratorAttachmentWorkItem] {
+      override def reads(json: JsValue): JsResult[NrsOrchestratorAttachmentWorkItem] =
+        Json.reads[NrsOrchestratorAttachmentWorkItem].reads(json) match {
```
josef-vlach: We should assign this to a `val`, see other comment for details.

```
Json.reads[NrsOrchestratorAttachmentWorkItem]
```

app/uk/gov/hmrc/gform/scheduler/nrsOrchestrator/NrsOrchestratorAttachmentWorkItem.scala
```diff
@@ -16,20 +16,72 @@
...
+                  workItem.attachment,
+                  workItem.businessId,
+                  workItem.notableEvent
+                )
+              )
+            } //TODO: This is a fallback for old workItem structure, remove this and NrsOrchestratorAttachmentWorkItemOld when database is cleaned-up.
+          case success: JsSuccess[NrsOrchestratorAttachmentWorkItem] => success
+        }
+      override def writes(o: NrsOrchestratorAttachmentWorkItem): JsValue =
+        Json.writes[NrsOrchestratorAttachmentWorkItem].writes(o)
```
josef-vlach: We should assign this to a `val`

```
Json.writes[NrsOrchestratorAttachmentWorkItem]
```

app/uk/gov/hmrc/gform/scheduler/nrsOrchestrator/NrsOrchestratorWorkItem.scala
```diff
@@ -36,8 +38,63 @@ final case class NrsOrchestratorWorkItem(
...
+                  workItem.payload,
+                  workItem.submissionDate,
+                  workItem.userAuthToken,
+                  workItem.identityData
+                )
+              )
+            } //TODO: This is a fallback for old workItem structure, remove this and NrsOrchestratorWorkItemOld when database is cleaned-up.
+          case success: JsSuccess[NrsOrchestratorWorkItem] => success
+        }
+      override def writes(o: NrsOrchestratorWorkItem): JsValue = Json.writes[NrsOrchestratorWorkItem].writes(o)
```
josef-vlach: We should assign this to a `val`

```
Json.writes[NrsOrchestratorWorkItem]
```

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/scheduler/nrsOrchestrator/NrsOrchestratorWorkItem.scala
PavelVjalicin: I thought about replacing "def" with "val". From what I understand macros are all compiled during scala compile time. So I don't think there is performance loss there, but I will change to "val" just in case.

By the time the "def" is being called the macro is already compiled, I don't think it would be even possible for macro to compile during runtime. It lacks necessary compile time context.

josef-vlach: Ok, we can keep it as it is, since it is a temporal code only anyway. 

Let's just remember to drop this fallback solution once old data structure support is not needed anymore.

josef-vlach (approved):

PR #1513, opened 2026-05-27
GFORMS-3689 - Replay the URI used in API calls within destinations/dataRetrieves

cmanson:


app/uk/gov/hmrc/gform/sharedmodel/DataRetrieveDefinition.scala
```diff
@@ -17,22 +17,53 @@
...
+}
+
 final case class DataRetrieveDefinition(
   tpe: DataRetrieve.Type,
   attributes: Attr,
   parameters: List[DataRetrieve.Parameter],
   attrTypeMapping: Map[DataRetrieve.Attribute, DataRetrieve.AttrType],
-  documentationUrl: Option[String]
+  documentationUrl: Option[String],
+  urlDescriptors: List[UrlDescriptor]
```
josef-vlach, line 70: Why is this a list?

Unless I'm missing something and this is supposed to be `UrlDescriptor` we should use it our Connectors, so we can guarantee, that this is indeed the url being used.

cmanson: It's a list because many of our data retrieve calls have 2 hops... 1 to gform backend then 1 to somewhere else.

cmanson: Regarding using connectors - many data retrieves are called from the frontend so we don't have access to those. This is just helpful documentation for anyone who needs to search for the URLs in Kibana. I get it that they could become out of date if a URL changes and the documentation isn't updated - however this is no different to any other documentation.

PR #1538, opened 2026-07-24
GFORMS-4630 - Accessibility ITST45 - Name fields missing autocomplete

cmanson:


conf/formTemplateSchema.json
```diff
@@ -815,6 +815,9 @@
         "hint": {
           "$ref": "#/$defs/stringOrEnCyObject"
         },
+        "autocomplete": {
```
josef-vlach, line 818: There is missing "rule" that `autocomplete` is only possible to define on "text" components.

josef-vlach (changes requested):
Next to missing json schema validation rule, see other comment, we should also allow "autocomplete" to be defined only on text component with "format": "text".

josef-vlach (approved):

PR #1541, opened 2026-07-31
GFORMS-4617: Slow queries in production

PavelVjalicin:
This should fix all slow queries that have been issued in the last 30 days.

Added indexes to sdesDestination and forms collections.
The build process of indexes might be resource intensive, should probably be deployed during low traffic time.

### DIFF
--- a/app/uk/gov/hmrc/gform/ApplicationLoader.scala
+++ b/app/uk/gov/hmrc/gform/ApplicationLoader.scala
@@ -239,18 +239,78 @@ class ApplicationModule(context: Context)
   private val prodCreatedExpiryDays: Int = configModule.appConfig.formExpiryDaysFromCreation
   private val prodSubmittedExpiryHours: Int = configModule.appConfig.submittedFormExpiryHours
   private val formsCacheRepository =
-    createMongoCacheRepository("forms", prodExpiryDays, prodCreatedExpiryDays, prodSubmittedExpiryHours)
+    createMongoCacheRepository(
+      "forms",
+      prodExpiryDays,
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -296,6 +296,18 @@ class ApplicationModule(context: Context)
           IndexOptions()
             .background(false)
             .name("statusLastUpdatedIdx")
+        ),
+        IndexModel(
+          ascending("data.form.userId"),
```
josef-vlach: This is a unique field. Why we need an index for it?

app/uk/gov/hmrc/gform/ApplicationLoader.scala
```diff
@@ -296,6 +296,18 @@ class ApplicationModule(context: Context)
...
+          IndexOptions()
+            .background(false)
+            .name("userIdIdx")
+        ),
+        IndexModel(
+          ascending("data.form.envelopeId"),
```
josef-vlach: This field doesn't exist on `snapshots` collection. It seems we will need to separate indexes into common and unique.

josef-vlach (changes requested):

app/uk/gov/hmrc/gform/ApplicationLoader.scala
PavelVjalicin: Did not realise that snapshots collection is also affected. I separated which indexes are used for snapshots and forms.

Did not intend to add any indexes to snapshots.

PR #1545, opened 2026-08-06
GFORMS-4642 - Reject json template with invalid field reference in emailParameters

cmanson:


josef-vlach (changes requested):
This is not the idiomatic solution.

Better solution is to define missing implicit 

```
implicit val leafEmailParameter: LeafExpr[EmailParameter] = (path: TemplatePath, t: EmailParameter) =>
    LeafExpr(path + "value", t.value)
```

and to modify `implicit val leafExprs: LeafExpr[FormTemplate]` by adding

```
LeafExpr(path + "emailParameters", t.emailParameters) ++
```

Then whole `def validateEmailParameter`  can be deleted.

josef-vlach (approved):

